PR #23, opened 2014-06-03
man page: add docs for two options from readme

vcunat:
Reported on ML by tjernlund@tjernlund.se.


vcunat:
Ping @edolstra or whoever has commit access.

edolstra:
Applied, thanks!

PR #26, opened 2014-07-23
Add '--remove-rpath’ option — removes sections instead of setting to empty string.

hmage:
Instead of setting to empty string, it removes both DT_RUNPATH and DT_RPATH sections — this makes ld.so's self checks happy.


edolstra:
Closing due to #50.

PR #51, opened 2015-07-14
Change configure.ac from OpenWrt to support Mac OS

pretyman:
This change is from the OpenWrt project, from the URL:
https://dev.openwrt.org/ticket/18998
It enables patchelf to sucessfully compile and run under Mac OS and
keeps compatibility with Linux.

Tested under Mac OS X 10.9.5 and Ubuntu 15.04 to patch a binary of type
ELF 32-bit LSB executable - the patch fully worked

This patch is credited to Felix Fietkau


edolstra:
Thanks, applied!

Issue #53, opened 2015-07-18
patchelf tests fail to pass on powerpc64


adevress:
The patchelf "make check" tests fail to execute on linux powerpc64.

PASS: set-interpreter-short.sh
FAIL: set-interpreter-long.sh
FAIL: set-rpath.sh
FAIL: no-rpath.sh
FAIL: big-dynstr.sh
FAIL: set-rpath-library.sh
PASS: soname.sh

This is a blocking issue for the usage of Nix/NixOS on the power architecture.

adevress:
This is not only a false positive.

It seems that patchelf corrupt binary under powerpc64 :

```
$ ./ls: /nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-glibc-2.21-powerpc64-unknown-linux-gnu/lib/ld64.so.1: bad ELF interpreter: Aucun fichier ou dossier de ce type

# patchelf to set the interpreter to a valid value
~/scratch/workspace/patchelf/src/patchelf --set-interpreter /gpfs/[.....]/nix/store/ng8p3bpwxlwp8dslv31dndkk4wrg79bp-bootstrap-tools/lib/ld-2.21.so ls

# retry execution
./ls
Processus arrêté

# valgrind trace 
 valgrind --tool=memcheck ./ls
==27375== Memcheck, a memory error detector
==27375== Copyright (C) 2002-2012, and GNU GPL'd, by Julian Seward et al.
==27375== Using Valgrind-3.8.1 and LibVEX; rerun with -h for copyright info
==27375== Command: ./ls
==27375== 
Inconsistency detected by ld.so: rtld.c: 1141: dl_main: Assertion `_rtld_local._dl_rtld_map.l_libname' failed!
==27375== 
==27375== HEAP SUMMARY:

**Comment by adevress** [comment]
Anyone has seen something similar this before ?

**Comment by adevress** [comment]
I add you a little over trace on this when using only native system tools

```
# copy ls in current directory
cp /bin/ls ./
# use it
./ls 
# -> works with normal listing
# use patch elf to print interpreter
 ~/scratch/workspace/patchelf/src/patchelf --print-interpreter --debug  ./ls
/lib64/ld64.so.1
# list again
./ls 
# works
# now modify the interpreter with exactly the same value than before
 ~/scratch/workspace/patchelf/src/patchelf --set-interpreter /lib64/ld64.so.1 --debug  ./ls
patching ELF file `ls'
[...]

PR #54, opened 2015-07-20
Solve issue #53 by getting the kernel page size directly at runtime

adevress:
This pull request solves the issues of patchelf on powerpc64 and improve protability.

Issue reported here :
https://github.com/NixOS/patchelf/issues/53

powerpc64 architectures may have a page size 65K instead of the usual 4K. 
This causes patchelf to corrupt any powerpc64 binary it touches.

This patch solve this problem by using sysconf() when available to determine the page size.


adevress:
No interest on this ?

edolstra:
Applied, thanks!

PR #56, opened 2015-07-24
Fix converting DT_RUNPATH to DT_RPATH

darealshinji:
Otherwise if I change a runpath and use `--force-rpath` it will still remain a DT_RUNPATH entry and not be converted to DT_RPATH.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1159,8 +1159,10 @@ void ElfFile<ElfFileParamNames>::modifyRPath(RPathOp op, string newRPath)
         dynRPath = 0;
     }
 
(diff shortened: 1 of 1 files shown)


darealshinji:
No interest?

darealshinji:
Any reason why this wasn't merged yet? It's a very minor change but it fixes an existing bug.

speleo3:
The description sounds like it would fix #94, but it doesn't. My [test script](https://github.com/NixOS/patchelf/issues/94#issuecomment-338183814) produces the same results.

darealshinji:
Closing in favor of https://github.com/NixOS/patchelf/pull/137

PR #57, opened 2015-07-24
Add --print-needed command to list all DT_NEEDED entries of a file

darealshinji:


src/patchelf.cc
```diff
@@ -1421,7 +1442,7 @@ static void patchElf2(ElfFile & elfFile, mode_t fileMode)
 
 static void patchElf()
 {
-    if (!printInterpreter && !printRPath && !printSoname)
+    if (!printInterpreter && !printRPath && !printSoname && !printNeeded)
```
fsateler: This seem incorrect. Printing the needed section will not avoid manupilating the file.

darealshinji: Wouldn't that mean it's incorrect for the other --print-* commands too?

fsateler: Yes, I think so.

PR #58, opened 2015-07-24
Prefer the system's elf header over the bundled one

darealshinji:
Usually packagers prefer to use the distribution's header file. However this way it should still build on non-ELF systems.


fsateler:
Yes please

Mic92:
Thanks for your contribution. For simplicity I would prefer to have a single source of truth.
The constants we use from `elf.h` are universally true.
I don't see any advantage using the systems version as the result will be the same.
If we start using the systems elf file we might break silently the build in case there is no `elf.h`
because we might accidentally use definitions not found in our vendored version.

PR #73, opened 2016-01-15
Allow specifying pagesize

fsateler:
This PR uses getconf to select the default pagesize (with a fallback to 4096). It also adds an option to specify the pagesize at runtime to operate on foreign binaries.

This is more of a RFC and call for tests, so please test!

/cc @adevress @domenkozar


adevress:
+1 for this patch, we need it for powerpc64 integration

edolstra:
Thanks, merged!

Issue #79, opened 2016-02-17
fail aarch64 -- Need to detect page size from ELF file


gitfoxi:
```
make check
Making check in src
make[1]: Entering directory `/home/admin/nix-boot/patchelf-0.8/src'
make[1]: Nothing to be done for `check'.
make[1]: Leaving directory `/home/admin/nix-boot/patchelf-0.8/src'
Making check in tests
make[1]: Entering directory `/home/admin/nix-boot/patchelf-0.8/tests'
make  simple main main-scoped big-dynstr libfoo.so libfoo-scoped.so libbar.so libbar-scoped.so
make[2]: Entering directory `/home/admin/nix-boot/patchelf-0.8/tests'
make[2]: `simple' is up to date.
make[2]: `main' is up to date.
make[2]: `main-scoped' is up to date.
make[2]: `big-dynstr' is up to date.
make[2]: `libfoo.so' is up to date.
make[2]: `libfoo-scoped.so' is up to date.
make[2]: `libbar.so' is up to date.
make[2]: `libbar-scoped.so' is up to date.
make[2]: Leaving directory `/home/admin/nix-boot/patchelf-0.8/tests'
make  check-TESTS
make[2]: Entering directory `/home/admin/nix-boot/patchelf-0.8/tests'
make[3]: Entering directory `/home/admin/nix-boot/patchelf-0.8/tests'
PASS: plain-fail.sh
PASS: plain-run.sh
PASS: shrink-rpath.sh
PASS: set-interpreter-short.sh
FAIL: set-interpreter-long.sh
PASS: no-rpath.sh
FAIL: set-rpath.sh
FAIL: big-dynstr.sh
FAIL: set-rpath-library.sh
make[4]: Entering directory `/home/admin/nix-boot/patchelf-0.8/tests'
make[4]: Nothing to be done for `all'.
make[4]: Leaving directory `/home/admin/nix-boot/patchelf-0.8/tests'
============================================================================
Testsuite summary for patchelf 0.8
============================================================================
# TOTAL: 9
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  4
# XPASS: 0
# ERROR: 0
============================================================================
See tests/test-suite.log
============================================================================
make[3]: *** [test-suite.log] Error 1
make[3]: Leaving directory `/home/admin/nix-boot/patchelf-0.8/tests'
make[2]: *** [check-TESTS] Error 2
[...]

PR #80, opened 2016-02-28
aarch64 page-size support

gitfoxi:
This isn't ready. It works natively on aarch64 patching elfs for same, but fails on cross-architecture patch. See issue #79


gitfoxi:
Passing all tests run under amd64 and aarch64.

Not sure what this "conflicts that must be resolved" thing is about.

fsateler:
@gitfoxi it means that the recent changes to the git repository mean that you changes no longer apply cleanly. You need to update the upstream repository and rebase on top of or merge the latest git master.

fsateler:
In particular, your code probably conflicts with pr #73

gitfoxi:
Thanks @fsateler 

I see that was a different solution to the same problem. Just let the user specify the page size. I thought of this but wondered how nixpkgs could feed the target page size to all of patchelf for a given platform and gave up.

My approach is to look at the architecture of the ELF being modified and use that to select page size. I think it's better but not perfect because, for example, aarch64 can have different page sizes, configured at the OS level. But 64k is a safe superset of all possible page sizes.

Ideally, the page size would be stored in the ELF file itself but that doesn't seem to be the case.

What do you think @fsateler? I don't really know how or if I should proceed.

fsateler:
@gitfoxi The page size is auto-detected at build time. This means that when running natively (ie, the same env that produced patchelf), the default should be correct. The page size is for operating on foreign binaries: those not generated by the same toolchain as your own patchelf binary.

Now, if you could somehow determine the page size by inspecting the elf binary, that would be awesome. As you note, the info is not readily available, so some gymnastics might be needed.

domenkozar:
See https://github.com/NixOS/patchelf/pull/216

Issue #84, opened 2016-04-01
patchelf --replace-needed forgets to modify .gnu.version_r section


njsmith:
If I have an ELF binary that is linked to a DSO that uses symbol versioning, then the DSO's SONAME appears in two different places: once as a DT_NEEDED entry, and once in the `.gnu.version_r` version requirements section. If I use `patchelf --replace-needed`, then it updates the DT_NEEDED entry, but doesn't update the `.gnu.version_r` table. The resulting binary is completely broken -- trying to load it triggers an assertion failure inside the dynamic loader, as it tries to check the version of a library that was never loaded:
```
Inconsistency detected by ld.so: dl-version.c: 224: _dl_check_map_versions: Assertion `needed != ((void *)0)' failed!
```

For example, here's readelf output for a library that's had `patchelf --replace-needed libgfortran.so.3 libgfortran-ed201abd.so.3.0.0` run on it:
```
$ readelf -a lib/python2.7/site-packages/numpy/.libs/libopenblasp-r0-4830d766.2.17.so
[...]
Dynamic section at offset 0x2318100 contains 28 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libgfortran-ed201abd.so.3.0.0]
[...]
Version needs section '.gnu.version_r' contains 4 entries:
 Addr: 0x00000000000aff18  Offset: 0x0aff18  Link: 28 (.dynstr)
  000000: Version: 1  File: libgfortran.so.3  Cnt: 1
  0x0010:   Name: GFORTRAN_1.0  Flags: none  Version: 7
```
Notice that the DT_NEEDED entry is correct, but the `.gnu.version_r` entry still refers to the original `libgfortran.so.3`.

The fix is simple, at least conceptually :-). `patchelf --replace-needed` should check for any references to the replaced SONAME inside the `.gnu.version_r` section and update it if necessary.

[...]

PR #85, opened 2016-04-01
Teach --replace-needed to update .gnu.version_r table

njsmith:
If the ELF binary that we're patching is linked to a DSO that uses
symbol versioning, then the DSO's SONAME appears in two different
places: once as a DT_NEEDED entry, and once in the .gnu.version_r
version requirements section. Previously, `patchelf --replace-needed`
would update DT_NEEDED entry, but fail to update the .gnu.version_r
table. This resulted in completely broken binaries -- trying to load
them would trigger an assertion failure inside the dynamic loader, as it
tries to check the version of a library that was never loaded:
```
Inconsistency detected by ld.so: dl-version.c: 224: _dl_check_map_versions: Assertion `needed != ((void *)0)' failed!
```
This PR teaches `--replace-needed` to update the .gnu.version_r
table.

It also contains a few unrelated cleanups to `replaceNeeded` in a separate commit.

There currently isn't any test included, because I couldn't find an existing test for `--replace-needed` so I guess that might be okay, plus I don't really understand the automake test framework and I need to go to sleep :-). If you want a test then I can probably figure it out, but would appreciate any hints...

Fixes: gh-84

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -53,8 +53,8 @@ off_t fileSize, maxSize;
 unsigned char * contents = 0;
 
 
-#define ElfFileParams class Elf_Ehdr, class Elf_Phdr, class Elf_Shdr, class Elf_Addr, class Elf_Off, class Elf_Dyn, class Elf_Sym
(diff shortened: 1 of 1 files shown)


edolstra:
Thanks, merged!

qqshka:
Sorry for necro bumping but '--remove-needed' have exactly the same issue.

PR #86, opened 2016-04-02
Fix bug in walking .gnu.version_r linked list

njsmith:
When writing the code to teach --replace-needed to modify the
.gnu.version_r section (gh-85), I misunderstood how the ->vn_next
pointers in the Elf_Verneed structs are supposed to be interpreted: I
thought they gave an offset from the beginning of the section, but in
fact they give an offset relative to the current struct. The resulting
bug was very odd: generally, patchelf would complete without signalling
an error, but it would only successfully replace filenames that occurred
as either the first or second entries in the .gnu.version_r section,
while the third or later entries would be left untouched.

This commit fixes the interpretation of the ->vn_next pointers, so that
now --replace-needed should work correctly even on ELF files with more
than two version needed structs.

Thanks to @matthew-brett for finding the bug / providing a test case,
and to @rmcgibbo for helping me diagnose it.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1337,7 +1337,7 @@ void ElfFile<ElfFileParamNames>::replaceNeeded(map<string, string>& libs)
                 debug("keeping .gnu.version_r entry `%s'\n", file);
             }
             // the Elf_Verneed structures form a linked list, so jump to next entry
-            need = (Elf_Verneed *) (contents + rdi(shdrVersionR.sh_offset) + rdi(need->vn_next));
+            need = (Elf_Verneed *) (((char *) need) + rdi(need->vn_next));
             --verNeedNum;
         }
     }


cstrahan:
@njsmith Awesome work! You're work on improving the correctness of `patchelf` is greatly appreciated. Bugs aside, I hope it's proven useful in wheels.

njsmith:
@cstrahan: thanks! And yes, it's proven *very* useful; it's an integral part of the [auditwheel] (https://pypi.python.org/pypi/auditwheel) tool that we use to make wheels self contained, and we probably couldn't practically make Linux wheels work without it. In fact I like it so much that I wrote my own version for PE files :-) (https://github.com/njsmith/redll)

njsmith:
Ping. We've been shipping this in the [manylinux docker image](https://github.com/pypa/manylinux) for a few weeks now and it seems to be working well.

cstrahan:
@edolstra Can we merge this?

edolstra:
Thanks, merged!

domenkozar:
Really glad Python community is slowly moving to Nix :-)

PR #89, opened 2016-04-04
handle read-only files

darealshinji:
Fixes an issue mentioned in nixpkgs: https://github.com/NixOS/nixpkgs/issues/14440


edolstra:
Arguably, if files are read-only, they are read-only for a reason, so programs should *not* chmod them on their own. Instead it should be the user's responsibility to make the file writable first.

darealshinji:
Good point. That's also how it was fixed in the .nix file.

darealshinji:
I'll close this PR, okay?

PR #101, opened 2016-06-13
add --shrink-prefix used for build while package is not installed

### LINKED ISSUE #100: add search path prefix for "--shrink-rpath"
In rpmbuild, software was built in builtroot some path like /tmp/buildroot/, and software install to real path after packaging. 

So I need to specify patchelf to search libs using prefix /tmp/buildroot

linxiulei:
FIX #100.

In addition, the '--shrink-prefix' option can be used for specify rpath
when executable is building in anothen buildroot. For instance, if an
executable located in /tmp/buildroot and can only be installed to
/usr/local. To accomplish that in build stage, use:

$ patchelf --shrink-rpath --shrink-prefix /tmp/buildroot

### DIFF
--- a/patchelf.1
+++ b/patchelf.1
@@ -53,6 +53,9 @@ For instance, if an executable references one library libfoo.so, has
 an RPATH "/lib:/usr/lib:/foo/lib", and libfoo.so can only be found
 in /foo/lib, then the new RPATH will be "/foo/lib".
 
+.IP --shrink-prefix
+Work with --shrink-rpath, used in stage of build before installed to real dest path.
+
 .IP --print-rpath
 Prints the RPATH for an executable or library.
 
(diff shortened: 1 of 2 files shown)


patchelf.1
```diff
@@ -53,6 +53,9 @@ For instance, if an executable references one library libfoo.so, has
 an RPATH "/lib:/usr/lib:/foo/lib", and libfoo.so can only be found
 in /foo/lib, then the new RPATH will be "/foo/lib".
 
+.IP --shrink-prefix
+Work with --shrink-rpath, used in stage of build before installed to real dest path.
```
edolstra: I can't really make sense of this description...

linxiulei: Thanks for response

I use patchelf in a stage of rpmbuild, which install software files in a buildroot dir ( for example /home/rpmbuild/buildroot but which is not the root dir software actually run at ).
And in that case, patchelf would not find dynamic libs correctly. 

.IP --shrink-prefix PREFIX
Work with --shrink-rpath, search all directory in RPATH with prefix $PREFIX, insteal search all directory IN RPATH

For instance as above,  --shrink-prefix /root/build/,  /root/build/lib:/root/build/usr/lib:/root/build/foo/lib would be searched

PR #104, opened 2016-07-10
tests: Do not exit on an expected failure code.

mkoloberdin:
The comment in the code below the patched line
clearly suggests that it should go on in this case.
This condition occurs on Alpine Linux (musl-libc-based):
http://patchwork.alpinelinux.org/patch/2206/


domenkozar:
Could you rebase?

domenkozar:
I've tested it and it works. Thanks!

PR #106, opened 2016-09-28
emit an error if multiple target files are given

benjaminp:
Only one file may be patched at a time. patchelf previously silently ignored extra command line arguments. Make extra arguments be an error.


fsateler:
wouldn't it be better to actually process the multiple files?

benjaminp:
Yes, but erroring out is better than silently ignoring extra files.

domenkozar:
Implemented in https://github.com/NixOS/patchelf/pull/123, can you test?

benjaminp:
cool

PR #109, opened 2016-12-03
added cmake build

linux-fan-dave:


peterhoeg:
Hi @linux-fan-dave, do you mind sharing some details regarding the purpose of this PR?

probonopd:
Looks like @linux-fan-dave prefers CMake over autotools, and I share this preference.

linux-fan-dave:
Ups sorry. I thought I did provide a description. I simply added a CMake build file as a alternative to the autotools build. If you already use CMake it is really easy to add dependencies to other CMake projects. Autotools do not integrate quite as nice.

fsateler:
This cmake script is not equivalent to the autotools one. Differences I can find:

1. Does not autodetect page size.
2. Does not include tests.
3. Enables C++11 mode
4. Does not install manpage

linux-fan-dave:
Would the CMake build file be accepted if it were without the mentioned problems? 
To 1:
What does autoedect page size mean? I saw a default in the configure and a paramter that can overwrite the auto. This is the behaviour in cmake as well. It has a default / auto and a paramter that can be used to overwrite.
To 2: 
Yes this is true. Could be done.
To 3: 
I was not able to compile without C++11 enabled - but I could investigate further.
To 4: 
I could add that.

fsateler:
@linux-fan-dave I am not a maintainer so I can't answer your overall question (but in general I like cmake more than autotools so +1 from me). Of course you should wait to see if the maintainer wants this change before doing more work on this.

As to 1, there is this line in configure that queries `getconf` to ask for the page size: https://github.com/NixOS/patchelf/blob/4034c6d/configure.ac#L17

Moreover, your parameter is not overridable in the command line. It should probably use the [CACHE option](https://cmake.org/cmake/help/v3.3/command/set.html#set-cache-entry), and provide a short help string.

vcunat:
> If you already use CMake it is really easy to add dependencies to other CMake projects. Autotools do not integrate quite as nice.

I don't get what dependencies you mean. Is there some other task than locating the `patchelf` executable? Surely `cmake` can handle that easily already.

slashbeast:
In general, cmake allows you to easly "add_subproject()" which is useful when you want to link (for example statically) against another library, then cmake will build another cmake project as dependency and provide you the thing.

Here however, patchelf is a application not library, so no much useful.

This said, CMake sucks much less than Autotools and I am honestly surprised when i see a software that is not from '90 but decide to start with autotools. If I could I would replace autotools everywhere with CMake. :)

vcunat:
*I'm getting off-topic: my experience is that autotools-based tend to be more robust &ndash; when you have some "unusual" demands, and I* was *surprised by that. In particular I dealt with wanting to put different parts of the application to different prefixes (`--libdir`, `--mandir`) which worked mostly flawlessly on more than a hundred nixpkgs packages, and I did some cross-compilation.*

linux-fan-dave:
In our case we have a cmake based project which needs to patch some binaries before beeing deployed. We therefore included patchelf as a third party project to our project and compile it. It therefore can be used in a postbuild step. 
A cmake based project can easily be added to our project, whilst using the autotools from our cmake build is a bit of a hassle. If there is any interest in adding it to patchelf - I would fix the problems with cmake to be added upstream to patchelf.

kangjoni76:
to detect automatically Linux PAGE_SIZE you could write cmake [execute_process!]( https://cmake.org/cmake/help/latest/command/execute_process.html) 

Something like this will do the trick
````
execute_process(COMMAND ${CMAKE_COMMAND}  getconf PAGESIZE OUTPUT_VARIABLE PAGE_SIZE_VAR) 

add_definition (-DPAGE_SIZE=${PAGE_SIZE_VAR})
````
But untested though...

DerDakon (changes requested):
This misses ```install()``` steps to actually place the compiled executable somewhere.

emlisa3162:
I'm going to see if I can get this working... fingers crossed

concatime:
I added Meson on [my fork](https://github.com/concatime/patchelf/tree/meson-build). I can create a PR if you are interested. It compiles successfully with [muon](https://sr.ht/~lattis/muon/) + [samurai](https://github.com/michaelforney/samurai) (neither Python nor a C++ compiler are required, only a C99 compiler).

Ericson2314:
Replaced by https://github.com/NixOS/patchelf/pull/610

Issue #102, opened 2016-06-14
make compiling without autotools easier


darealshinji:
I sometimes like to download the latest snapshot of patchelf.cc when I need it and compile it in a single command like this: `g++ -DPAGESIZE=4096 -DPACKAGE_STRING=\"\" -std=c++11 -o patchelf patchelf.cc`
It would be convenient if `PAGESIZE` and `PACKAGE_STRING` did fall back to some default values if not specified through CXXFLAGS, in such a way that it of course won't break the autotools build system. I'm thinking of adding something like this:

``` c++
#ifndef PAGESIZE
#  define PAGESIZE 4096
#endif
#ifndef PACKAGE_STRING
#  define PACKAGE_STRING "unknown"
#endif
```

Ericson2314:
This was fixed in $299

PR #114, opened 2017-01-30
always define PACKAGE_STRING and PAGESIZE

darealshinji:
This way you can compile the program in a simple one-liner. That can always come in handy.


domenkozar:
Fixes #102

Issue #10, opened 2014-01-17
Section re-ordering confuses `strip`


staticfloat:
Running `patchelf` to enlarge my RUNPATH entry causes my section headers to change from this:

```
Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .interp           PROGBITS         0000000000400238  00000238
       000000000000001c  0000000000000000   A       0     0     1
  [ 2] .note.ABI-tag     NOTE             0000000000400254  00000254
       0000000000000020  0000000000000000   A       0     0     4
  [ 3] .note.gnu.build-i NOTE             0000000000400274  00000274
       0000000000000024  0000000000000000   A       0     0     4
  [ 4] .gnu.hash         GNU_HASH         0000000000400298  00000298
       0000000000000064  0000000000000000   A       5     0     8
  [ 5] .dynsym           DYNSYM           0000000000400300  00000300
       00000000000011b8  0000000000000018   A       6     1     8
  [ 6] .dynstr           STRTAB           00000000004014b8  000014b8
       0000000000000868  0000000000000000   A       0     0     1
...
```

To this:

```
Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .dynsym           DYNSYM           00000000003ffb00  001ffb00
       00000000000011b8  0000000000000018   A       2     1     8
  [ 2] .dynstr           STRTAB           00000000003ff270  00000270
       0000000000000890  0000000000000000   A       0     0     8
  [ 3] .gnu.hash         GNU_HASH         0000000000400cb8  00001cb8
       0000000000000064  0000000000000000   A       1     0     8
  [ 4] .interp           PROGBITS         0000000000400d20  00001d20
       000000000000001c  0000000000000000   A       0     0     8
  [ 5] .note.ABI-tag     NOTE             0000000000400d40  00001d40
       0000000000000020  0000000000000000   A       0     0     8
  [ 6] .note.gnu.build-i NOTE             0000000000400d60  00001d60
       0000000000000024  0000000000000000   A       0     0     8
...
```

[...]

PR #117, opened 2017-03-07
Avoid inflating file sizes needlessly and allow binaries to be stripped

rpurdie:
The current approach to changing sections in ET_DYN executables is to move
the INTERP section to the end of the file. +This means changing PT_PHDR to
add an extra PT_LOAD section so that the new section is mmaped into memory
by the elf loader in the kernel. In order to extend PHDR, this means moving
it to the end of the file.

Its documented in BUGS there is a kernel 'bug' which means that if you have holes
in memory between the base load address and the PT_LOAD segment that contains PHDR,
it will pass an incorrect PHDR address to ld.so and fail to load the binary, segfaulting.

To avoid this, the code currently inserts space into the binary to ensure that when
loaded into memory there are no holes between the PT_LOAD sections. This inflates the
binaries by many MBs in some cases. Whilst we could make them sparse, there is a second
issue which is that strip can fail to process these binaries:

$ strip fixincl
Not enough room for program headers, try linking with -N
[.note.ABI-tag]: Bad value

This turns out to be due to libbfd not liking the relocated PHDR section either
(https://github.com/NixOS/patchelf/issues/10).

Instead this patch implements a different approach, leaving PHDR where it is but extending
it in place to allow addition of a new PT_LOAD section. This overwrites sections in the
binary but those get moved to the end of the file in the new PT_LOAD section.

This is based on patches linked from the above github issue, however whilst the idea
was good, the implementation wasn't correct and they've been rewritten here.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -170,6 +170,8 @@ private:
     std::string & replaceSection(const SectionName & sectionName,
         unsigned int size);
 
+    bool haveReplacedSection(const SectionName & sectionName);
+
     void writeReplacedSections(Elf_Off & curOff,
         Elf_Addr startAddr, Elf_Off startOffset);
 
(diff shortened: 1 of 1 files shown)


probonopd:
Fixing stripping would indeed be very valuable. Thanks!

domenkozar:
@edolstra

EvenOldridge:
I'd love to see this committed as well.  Right now the numpy build process uses this and stripping binaries is essential for aws lambda deployment.

edolstra:
Thanks, merged!

zeerd:
dose this issue closed?

I met this today in the 0.9 version.

arch:x86_64

rpurdie:
There was a bug in my change which https://github.com/NixOS/patchelf/pull/127 addresses and may help.

PR #120, opened 2017-04-16
Refactor by extracting a couple o helper unctions

dezgeg:
Extract a couple of recurring patterns to their own helper functions.

This currently shouldn't cause any change in behaviour, but in the future I plan to hook `sectionContents()` to access `replacedSections`. This should fix cases where multiple operations are passed in (e.g. `--replace-needed foo bar --add-needed baz`). Currently, both of those do their work on the original file contents so the end result is some weird mixture of both actions.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -168,6 +168,8 @@ private:
 
     unsigned int findSection3(const SectionName & sectionName);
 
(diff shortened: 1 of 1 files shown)


domenkozar:
@dezgeg could you rebase? :/

Mic92:
Please re-open if you are still interested in this.

PR #121, opened 2017-04-19
Better error messages when run on statically-linked (or otherwise weird) binaries

dezgeg:
Make patchelf give nicer error messages. In particular, some UPX-compressed files in the wild have no section header table at all, which currently prints a `Assertion 'shstrtabIndex < shdrs.size()' failed.` message.


domenkozar:
Good way to test this is: http://fw.withings.net/pairingwizard_Linux_x86_64

PR #123, opened 2017-05-28
Allow multiple filenames to patch

lheckemann:
This makes behaviour less confusing when multiple filenames are
passed — previously, any extra filenames would be ignored completely,
as would any options passed after a filename. Now these are taken
into account.


lheckemann:
@edolstra ping

dtzWill:
I've been burned by this before, :+1: to this (or throwing an error).

edolstra:
Thanks!

PR #129, opened 2017-08-01
configure.ac: use "command -v" instead of "type -p"

bjornfor:
"type -p" is a Bash thing, whereas "command -v" is specified by POSIX
(more portable).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -13,7 +13,7 @@ AC_ARG_WITH([page-size],
 )
 
 if test "$PAGESIZE" = auto; then
-    if type -p getconf &>/dev/null; then
+    if command -v getconf >/dev/null; then
         PAGESIZE=$(getconf PAGESIZE || getconf PAGE_SIZE)
     fi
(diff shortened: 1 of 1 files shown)


domenkozar:
Can you point to docs claiming that? :)

bjornfor:
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/type.html

The first link shows that "command -v" is in POSIX and the latter that "type" (in POSIX) doesn't support any flags. "command -v" works in e.g. dash _and_ bash, whereas "type -p" only works in bash.

bjornfor:
Some practical examples:

```
$ bash -c "type -p getconf"
/run/current-system/sw/bin/getconf
$ bash -c "command -v getconf"
/run/current-system/sw/bin/getconf
$ dash -c "type -p getconf"
-p: not found
getconf is /run/current-system/sw/bin/getconf
### the above command exited with status 127
$ dash -c "command -v getconf"
/run/current-system/sw/bin/getconf
```

PR #137, opened 2017-10-20
Consistently use --force-rpath

### LINKED ISSUE #94: Cannot force rpath
I'm having trouble setting the `RPATH` on a binary when it already has a `RUNPATH`, even when the `--force-rpath` option is employed. Test case:

```
cp `which ls` .
readelf -d ./ls | egrep -i "runpath|rpath"
```

Displays nothing, as ls by default appears to have no rpath or runpath.

```
patchelf --set-rpath ~/libs ./ls
readelf -d ./ls | egrep -i "runpath|rpath"
```

Now it displays `0x000000000000001d (RUNPATH)            Library runpath: [/home/mpn/libs]`. OK, fine!

Next do:

```
patchelf --force-rpath --set-rpath ~/libs ./ls
readelf -d ./ls | egrep -i "runpath|rpath"
```

Expected result: It will display an RPATH for the binary
Actual result: It still displays only a RUNPATH for the binary

My current workaround has been to delete the existing RUNPATH using `chrpath -d` (using --set-rpath "" with patchelf doesn't appear to do it, as that looks rather to set RUNPATH to an empty string instead), and _then_ to use `--force-rpath` with patchelf. Force-rpath does appear to work when there is no RUNPATH present.

speleo3:
Fixes #94

Produces expected results with [my test script](https://github.com/NixOS/patchelf/issues/94#issuecomment-338183814).

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1259,7 +1259,17 @@ void ElfFile<ElfFileParamNames>::modifyRPath(RPathOp op,
     }
 
 
-    if (std::string(rpath ? rpath : "") == newRPath) return;
+    if (!forceRPath && dynRPath && !dynRunPath) { /* convert DT_RPATH to DT_RUNPATH */
+        dynRPath->d_tag = DT_RUNPATH;
+        dynRunPath = dynRPath;
+        dynRPath = 0;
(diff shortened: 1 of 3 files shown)


domenkozar:
@speleo3 could you provide a test case in tests directory?

speleo3:
@domenkozar I've added a test.

PR #139, opened 2017-12-07
Bugfix for CompPhdr::operator() that could cause sort to fail.

kmillar:
Modified CompPhdr::operator() so that it provides a strict weak ordering as required by std::sort.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -135,7 +135,10 @@ private:
         ElfFile * elfFile;
         bool operator ()(const Elf_Phdr & x, const Elf_Phdr & y)
         {
-            if (x.p_type == PT_PHDR) return true;
+            if (x.p_type == PT_PHDR) {
+                  if (y.p_type == PT_PHDR) return false;
+                  return true;
+            }
             if (y.p_type == PT_PHDR) return false;
             return elfFile->rdi(x.p_paddr) < elfFile->rdi(y.p_paddr);
         }
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -135,12 +135,12 @@ private:
         ElfFile * elfFile;
         bool operator ()(const Elf_Phdr & x, const Elf_Phdr & y)
         {
-            if (x.p_type == PT_PHDR) {
-                  if (y.p_type == PT_PHDR) return false;
-                  return true;
-            }
-            if (y.p_type == PT_PHDR) return false;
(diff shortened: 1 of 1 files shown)


domenkozar:
Do you have a test case?

src/patchelf.cc
```diff
@@ -135,7 +135,10 @@ class ElfFile
         ElfFile * elfFile;
         bool operator ()(const Elf_Phdr & x, const Elf_Phdr & y)
         {
-            if (x.p_type == PT_PHDR) return true;
+            if (x.p_type == PT_PHDR) {
+                  if (y.p_type == PT_PHDR) return false;
+                  return true;
+            }
```
dtzWill: This could be simplified a bit (mostly removing the duplicated check of `y.p_type == PT_PHDR`:

```
// Nothing is less than a PHDR:
if (y.p_type == PT_PHDR) return false;

// PHDR's are less than everything (that's not a PHDR)
if (x.p_type == PT_PHDR) return true;

// Sort non-PHDR's by address:
...
```

Comments are just suggestions.

dtzWill: While it does seem worth fixing this, according to documents such as [1][2] it seems that only a single `PT_PHDR` entry may legally exist in a file.

Thinking about what `PT_PHDR` actually means, this makes sense :). Also my comments are indeed a bit silly, I didn't have my "ELF-knowledge hat" on.

So this is probably harmless as-is: for any legal set of segments this operator will define a valid (strict weak) ordering.

Might be worth ensuring we reject files that attempt to have multiple PT_PHDR segments?

[1] http://www.sco.com/developers/gabi/latest/ch5.pheader.html
[2] http://www.skyfree.org/linux/references/ELF_Format.pdf

dtzWill:
@domenkozar see my comment on the commit -- I'm not sure a test case (as an ELF input file) could be produced with normal tools.  Even if such a file were constructed, `patchelf` likely wouldn't do anything observably incorrect, unless running with some sort of debug version of `std::sort`?

Although if @kmillar has such a file that certainly would be interesting! :)

src/patchelf.cc
kmillar: I've simplified the code now.  Thanks.

Yes, this was found using a debug version of std::sort, not by obviously incorrect results.  Since there's only a single PT_PHDR entry, It probably doesn't any difference on any sane implementation of std::sort, but it is undefined behavior, so there's no guarantees about that.

Issue #66, opened 2015-09-16
patchelf --set-interpreter fails with `cannot find section `


pmahoney:
I have a binary compiled with the Go 1.4 compiler from `nixpkgs`, and I'd like to run it outside of a nix environment, so I'm trying to patch the interpreter.

patchelf 0.8 gives the following error: 

```
$ patchelf --set-interpreter /lib64/ld-linux-x86-64.so.2  prog
patchelf: patchelf.cc:693: void ElfFile<Elf_Ehdr, Elf_Phdr, Elf_Shdr, Elf_Addr, Elf_Off, Elf_Dyn, Elf_Sym>::rewriteSectionsExecutable() [with Elf_Ehdr = Elf64_Ehdr; Elf_Phdr = Elf64_Phdr; Elf_Shdr = Elf64_Shdr; Elf_Addr = long unsigned int; Elf_Off = long unsigned int; Elf_Dyn = Elf64_Dyn; Elf_Sym = Elf64_Sym]: Assertion `(off_t) rdi(hdr->e_shoff) >= startOffset' failed.
Aborted
```

Which I think was fixed in [pull 39](https://github.com/NixOS/patchelf/pull/39).

I built a patchelf from master @ 63296c4e18381216f740670916608f9ef159672e, and I now get a different error:

```
 ./result/bin/patchelf --set-interpreter /lib64/ld-linux-x86-64.so.2 --debug prog
patching ELF file `prog'
Kernel page size is 4096 bytes
replacing section `.interp' with size 28
this is an executable
using replaced section `.interp'
last replaced is 2
looking at section `'
replacing section `' which is in the way
looking at section `.interp'
first reserved offset/addr is 0xc00/0x400c00
first page is 0x400000
needed space is 600
clearing first 2504 bytes
cannot find section 
```

headers:

```
$ readelf -e prog
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              EXEC (Executable file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x43dee0
  Start of program headers:          64 (bytes into file)
  Start of section headers:          568 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
[...]

PR #149, opened 2018-06-21
Fix issue #66: ignore 0th section header when sorting, don't overwrite NOBITS

ezquat:
To recount the problem symptoms: Given an executable with certain properties, including a section at offset 0, patchelf fails to set the interpreter. (With patchelf 0.8 you get an assertion failure, (off_t) rdi(hdr->e_shoff) >= startOffset'. With patchelf 0.9 you get cannot find section, which is referring to the section with the empty name, which is some kind of dummy placeholder in ELF files, at section index 0.)

Here's my understanding of the bug: Most of the code in patchelf ignores the 0th entry in the section header table, but there is a call to C++ `sort()` which can move the entry at index 0, in the case where some other entry also has offset 0 (which is the case for Go binaries, a .tbss section has offset 0). It is easy to skip entry 0 in the sort. Then another problem kicks in, which is that patchelf tries to overwrite some bytes (with 'X's) at that offset 0, even though the `.tbss` section is marked as NOBITS and thus has no content.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -475,7 +475,7 @@ void ElfFile<ElfFileParamNames>::sortShdrs()
     /* Sort the sections by offset. */
     CompShdr comp;
     comp.elfFile = this;
-    sort(shdrs.begin(), shdrs.end(), comp);
+    sort(shdrs.begin() + 1, shdrs.end(), comp);
 
     /* Restore the sh_link mappings. */
     for (unsigned int i = 1; i < rdi(hdr->e_shnum); ++i)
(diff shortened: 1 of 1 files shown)


gilligan:
same here, i could really use this!

kyrofa:
Thank you @ezquat, this works for my test cases as well! Can we please get this merged?

Jumblemuddle:
+1 I'm unable to build the binaries for the [pulumi](https://github.com/pulumi/pulumi/) CLI.

kyrofa:
@edolstra, any thoughts on this?

sanmai-NL:
@puffnfresh: could you explain why this hasn't been merged yet?

theKAKAN:
> @puffnfresh: could you explain why this hasn't been merged yet?

He has approved the changes, it's @edolstra we are waiting for.

sjackman:
@edolstra @puffnfresh `patchelf` 0.9 was released three years ago. `master` includes a few helpful bug fixes. Are there any plans for a stable release?

offlinehacker:
Released in 0.10 for anyone wondering

karlyeurl:
Great! Thanks for the heads-up.

PR #151, opened 2018-06-25
Fix #150 - Check ELF endianness before writing new runpath

### LINKED ISSUE #150: --make-rpath-relative breaks PPC packages
When compiling some packages for PPC architecture, patchelf does not take into account the endianness (as far as I can tell) of the ELF file.  When cross compiling for PPC, the ELF file is broken after running patchelf with the --make-rpath-relative option.

The following entry below illustrates the issue:

0x1d000000 (<unknown>: 1d000000)        0xb77

When it should be:

0x0000001d (RUNPATH)                    Library runpath

In `patchelf.cc:1318` the `d_tag` field is being set to a hard define  `DT_RUNPATH`.  This appears to only be valid when host and target architectures are the same.

UraniumKnight:
This pull request makes a change to the way the RUNPATH field is updated in the dynamic section. Currently, patchelf does not take into account the target architecture of the ELF file before doing this (when --make-rpath-relative is specified).  Thus, the elf file ends up having an invalid tag for RUNPATH if the endianness differs. (0x1d000000 instead of 0x0000001d). 

The change proposed here updates the way these fields are written using the already defined endian-safe functions.  Tested on both PowerPC and ARM targets with a x86_64 host.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1269,13 +1269,13 @@ void ElfFile<ElfFileParamNames>::modifyRPath(RPathOp op,
     debug("new rpath is '%s'\n", newRPath.c_str());
 
     if (!forceRPath && dynRPath && !dynRunPath) { /* convert DT_RPATH to DT_RUNPATH */
-        dynRPath->d_tag = DT_RUNPATH;
+        wri(dynRPath->d_tag, DT_RUNPATH);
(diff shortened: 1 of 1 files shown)


domenkozar:
@UraniumKnight there are conflicts now, is this PR still needed on master?

Issue #44, opened 2014-11-09
ldconfig warns about truncated libraries


darealshinji:
ldconfig puts a warning on files that were modified with patchelf:

```
djcj patchelf-master $ sudo src/patchelf --set-rpath /usr/lib /usr/lib/x86_64-linux-gnu/libx264.so.142
djcj patchelf-master $ LANG=C sudo ldconfig
/sbin/ldconfig.real: file /usr/lib/x86_64-linux-gnu/libx264.so.142 is truncated

djcj patchelf-master $ 
```

AaronDMarasco-VSI:
I am getting the same error when using the `--set-soname` option on 63296c4e18381216f740670916608f9ef159672e.

sjackman:
I'm seeing this error as well. Do you have a fix or workaround?

AaronDMarasco-VSI:
For me with `--set-soname`, I am pretty sure `replaceSection()` is broken. Since I am always shortening the name, I have a custom patch that detects I am calling it and does the name replacement directly in memory by zeroing out the unused bytes at the end. The original code always assumed the new name was longer and called `replaceSection()`.

sjackman:
I use `--set-interpreter` and `--set-rpath` and they usually get longer for me. Thanks for the tip though. I could maybe arrange it for them to get shorter rather than longer. Probably not always though.

darealshinji:
@AaronDMarasco-VSI Can you provide said custom patch?

AaronDMarasco-VSI:
@darealshinji It doesn't fix `replaceSection()`; instead it bypasses it in a very specific way that would be pretty useless for most people. I compile the .so with the suffix "`_s`" but then later rename it to without so need to change that name explicitly.

Here's the [patch](https://gist.github.com/AaronDMarasco-VSI/cacfade609b77d8627567eaf3202d30e), but like I said, extremely specific and unlikely to be helpful for public consumption.

Narthorn:
It looks like ldconfig expects .dynstr to be in the first LOAD segment ([elf/readelf.c](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/readelflib.c;h=9ad56dcc341d86b975aa426b22de2ad31c830b94;hb=HEAD) in glibc), although I don't know why.

dandric-activevideo:
[...]

PR #159, opened 2018-09-27
Overwrite soname in place if there is room

reidpr:
This addresses issue #44 for the case where the new soname is the same or shorter than the old.

It's not great code. I'm not a C++ programmer, and I couldn't figure out how to make the overwrite idiomatic. So, it uses C string functions.

But, it does work for us.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1082,6 +1082,15 @@ void ElfFile<ElfFileParamNames>::modifySoname(sonameMode op, const std::string &
 
     debug("new SONAME is '%s'\n", newSoname.c_str());
 
+    /* If new SONAME fits in existing space, overwrite. This works around
+       issue #44. */
+    if (newSoname.size() <= sonameSize) {
+       debug("overwriting old SONAME with new...\n");
+       strcpy(soname, newSoname.c_str());
+       changed = true;
+       return;
+    }
+
     /* Grow the .dynstr section to make room for the new SONAME. */
     debug("SONAME is too long, resizing...\n");
 


haampie:
Any reason this was not merged?

domenkozar:
It would be great if you can create a regression test case in tests directory.

reidpr:
@domenkozar I agree 100%. Is there documentation on how to run the tests and how to write a new one? I didn't see anything at first glance. I don't have much time for this, and I apologize for that, but I don't see me adding a test unless it's very clear and straightforward.

Thanks for your hard work on patchelf! It's extremely useful for us.

domenkozar:
@reidpr see https://github.com/NixOS/patchelf/commit/81c0ea3eb87b077e0089e0f2737a63654e8bf51a

You can see a bunch of examples there, you'll also need to provide a sample binary to test.

reidpr:
OK, thanks for your help. `make check` I can handle but to write a test I need clear step-by-step instructions. I see there is a bunch of stuff in the `tests` directory, but I don't have time to reverse engineer the procedure to write a new test for an unfamiliar code base.

domenkozar:
@reidpr if you provide a binary and a small bash script to call patchelf and assert it works (and confirm it doesn't with the old patchelf) that should be sufficient.

src/patchelf.cc
```diff
@@ -1082,6 +1082,15 @@ void ElfFile<ElfFileParamNames>::modifySoname(sonameMode op, const std::string &
 
     debug("new SONAME is '%s'\n", newSoname.c_str());
 
+    /* If new SONAME fits in existing space, overwrite. This works around
+       issue #44. */
+    if (newSoname.size() <= sonameSize) {
+       debug("overwriting old SONAME with new...\n");
+       strcpy(soname, newSoname.c_str());
```
Mic92, line 1089: Since we already know the size of `newSoname`, memcpy is the better choice:

```suggestion
       memcpy(soname, newSoname.c_str(), newSoname.size() + 1);
```

src/patchelf.cc
```diff
@@ -1082,6 +1082,15 @@ void ElfFile<ElfFileParamNames>::modifySoname(sonameMode op, const std::string &
...
+       changed = true;
+       return;
```
fzakaria, line 1091: This functionality will have to change now @reidpr 
I introduced new changes that always do `this->rewriteSections();` after each `changed = true` This is to support multiple commands on a single invocation.

I agree that this is not an ideal tracking each `return`.
I would like to figure out how to do a destructor for each method and then apply `this->rewriteSections();` only if `changed == true`.

Either way this code will produce errors for now if done with other commands.

fzakaria (changes requested):
Please see my latest message.

src/patchelf.cc
fzakaria: I opened https://github.com/NixOS/patchelf/pull/363/files# to help make this simpler.

reidpr: Hello, thanks for the heads up. I still hope (after 3 years) this PR will get merged, but I don't have time to maintain it.

PR #166, opened 2019-03-30
Fixes for PIE executables

rcombs:
Most notably, this attempts to avoid growing the dynamic section when adding an rpath, since that results in repositioning the section, which means that existing IP-relative references to the `_DYNAMIC` symbol will become invalid.


domenkozar:
@rcombs could you rebase?

rcombs:
Done.

Mic92:
It seems that PRs with tests are preferred :) https://github.com/NixOS/patchelf/pull/225

src/patchelf.cc
```diff
@@ -741,10 +737,9 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
        since DYN executables tend to start at virtual address 0, so
        rewriteSectionsExecutable() won't work because it doesn't have
        any virtual address space to grow downwards into. */
-    if (isExecutable && startOffset > startPage) {
+    if (startOffset > startPage)
         debug("shifting new PT_LOAD segment by %d bytes to work around a Linux kernel bug\n", startOffset - startPage);
-        startPage = startOffset;
```
Mic92, line 746: This disables this mitigation?

Mic92: no, it unconditionally applies it...

src/patchelf.cc
```diff
@@ -741,10 +737,9 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
...
-    }
+    startPage = startOffset;
```
Mic92, line 742: I understand why `isExecutable` was removed as a conditional but not why `startPage = startOffset` was moved out.

Issue #155, opened 2018-08-08
--set-soname corrupts binary if the original soname string is also used for symbol versioning


mayeut:
Seen with NVIDIA CUDNN 7.1.4 for CUDA 9
Symbols are versioned using the SONAME i.e. libcudnn.so.7
When setting a new SONAME , all symbols become versioned with XXXXXXXXXXXXX

When commenting the following line, symbol version are correct.
https://github.com/NixOS/patchelf/blob/27ffe8ae871e7a186018d66020ef3f6162c12c69/src/patchelf.cc#L1080

I don't know the implication of commenting this line of code so not opening a PR and letting someone more enlightened on the subject follow-up on this.

Simple reproducer:
`lib.c`:
```
int exportedSymbol(int a) {
	return a;
}
```
`liblib.map`:
```
liblib.so.1 {
  global:
    exportedSymbol;

  local:
    *;
};
```
`script`:
```
gcc -shared -o liblib.so.1.0.0 -Wl,-soname,liblib.so.1 -Wl,--version-script=liblib.map lib.c
patchelf --set-soname liblib-abcd.so.1 liblib.so.1.0.0
```

This was tested in a quay.io/pypa/manylinux1_x86_64 docker container.

dezgeg:
This has happened before: https://github.com/NixOS/patchelf/issues/45#issuecomment-294801888

lkollar:
This is causing issues in `auditwheel` as reported in https://github.com/pypa/auditwheel/issues/159. Removing this "zeroing out" logic from `patchelf` fixes the issue. Does anybody know why this logic is necessary?

njsmith:
I think the memset is probably unnecessary, and only included to help catch any places where the soname was accidentally still used. Now we've discovered that it can be *intentionally* still used, so maybe it's best to remove the memset...

I guess the other option would be to explicitly check for any usage in the symbol table, and either rewrite those, or conditionally disable the memset only in cases where the soname is used as a version. But that's a lot more complicated than just deleting the memset.

PR #171, opened 2019-05-29
Remove zeroing out logic in modifySoname

lkollar:
The original SONAME is filled with 'X' characters in the modifySoname
function. This can cause issues if the .dynstr entry is still referenced
in other sections, e.g. some libraries use the SONAME entry as version
specifiers.

Closes: #155

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1074,13 +1074,6 @@ void ElfFile<ElfFileParamNames>::modifySoname(sonameMode op, const std::string &
         return;
     }
 
-    /* Zero out the previous SONAME */
-    unsigned int sonameSize = 0;
-    if (soname) {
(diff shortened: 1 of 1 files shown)


DelightRun:
Why has this PR still not been accepted?

domenkozar:
@lkollar I think I can help you get this merged if you provide a test case and a bit more explanations why that part of the code is not needed anymore.

domenkozar:
Seems like this was added in https://github.com/NixOS/patchelf/commit/764dacc6d6e54e343bbe62ea05e3df7dbd175e2c, @darealshinji could you take a look what was the reasoning for zeroing out soname?

domenkozar:
Given that this patch is widely used, let's just merge it.

Issue #82, opened 2016-03-01
Broken binaries generated on mipsel


fsateler:
Tests failed on mipsel on debian:

```
FAIL: set-rpath.sh
==================

Kernel page size is 16384 bytes
patching ELF file `scratch/set-rpath/main'
Kernel page size is 16384 bytes
new rpath is `/oops:/home/fsateler/patchelf/tests/scratch/set-rpath/libsA:/home/fsateler/patchelf/tests/scratch/set-rpath/libsB'
rpath is too long, resizing...
DT_NULL index is 27
replacing section `.dynamic' with size 272
replacing section `.dynstr' with size 404
this is an executable
using replaced section `.dynamic'
using replaced section `.dynstr'
last replaced is 9
looking at section `.interp'
replacing section `.interp' which is in the way
looking at section `.note.ABI-tag'
replacing section `.note.ABI-tag' which is in the way
looking at section `.MIPS.abiflags'
replacing section `.MIPS.abiflags' which is in the way
looking at section `.reginfo'
replacing section `.reginfo' which is in the way
looking at section `.note.gnu.build-id'
replacing section `.note.gnu.build-id' which is in the way
looking at section `.dynamic'
looking at section `.hash'
replacing section `.hash' which is in the way
looking at section `.dynsym'
replacing section `.dynsym' which is in the way
looking at section `.dynstr'
first reserved offset/addr is 0x65e/0x40065e
first page is 0x400000
needed space is 1748
needed space is 1780
needed pages is 1
changing alignment of program header 4 from 65536 to 16384
changing alignment of program header 5 from 65536 to 16384
clearing first 17578 bytes
rewriting section `.MIPS.abiflags' from offset 0x41c8 (size 24) to offset 0x1b4 (size 24)
rewriting section `.dynamic' from offset 0x421c (size 264) to offset 0x1cc (size 272)
rewriting section `.dynstr' from offset 0x453c (size 290) to offset 0x2dc (size 404)
rewriting section `.dynsym' from offset 0x43cc (size 368) to offset 0x470 (size 368)
rewriting section `.hash' from offset 0x4324 (size 168) to offset 0x5e0 (size 168)
rewriting section `.interp' from offset 0x4194 (size 13) to offset 0x688 (size 13)
rewriting section `.note.ABI-tag' from offset 0x41a4 (size 32) to offset 0x698 (size 32)
[...]

PR #180, opened 2019-08-22
Fixes for MIPS binaries

iv-m:
A couple of changes that address problems described in #82.

Tested on two machines, Baikal BE-T1000 and Loongson 3A, running mipsel port of ALT Linux Sisyphus. With these two changes, all tests pass on both of the machines, and on my x86_64 machine.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1098,9 +1098,9 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
        (e.g., those produced by klibc's klcc). */
     auto shdrDynamic = findSection2(".dynamic");
     if (shdrDynamic) {
-        auto dyn = (Elf_Dyn *)(contents + rdi(shdrDynamic->sh_offset));
+        auto dyn_table = (Elf_Dyn *) (contents + rdi(shdrDynamic->sh_offset));
         unsigned int d_tag;
-        for ( ; (d_tag = rdi(dyn->d_tag)) != DT_NULL; dyn++)
+        for (auto dyn = dyn_table; (d_tag = rdi(dyn->d_tag)) != DT_NULL; dyn++)
             if (d_tag == DT_STRTAB)
                 dyn->d_un.d_ptr = findSection(".dynstr").sh_addr;
             else if (d_tag == DT_STRSZ)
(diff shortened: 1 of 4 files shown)


iv-m:
UPD: replaced `ptrdiff_t`  with `auto` to make gcc5 happy (I was testing with gcc8).

olekw:
@edolstra: are you ok with this pull request? Unavailability of patchelf on MIPS is leading to a bunch of issues downstream in Debian. To expedite a fix, we could take this patch and implement it downstream until a newer version of patchelf is released. However, I'm hesitant to do that if you have a fundamental disagreement with how this solves #82. Thanks!

domenkozar:
@darealshinji would you review/test this?

fsateler:
I can confirm this makes tests pass on debian mipsel, on a Rhino Labs UTM8

domenkozar:
It would be great if someone stepped up to be maintainer of MIPS as per #199

iv-m:
UPD: rebased onto the current master with a couple of minor cosmetic changes.

src/patchelf.cc
```diff
@@ -1129,6 +1141,14 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
                 dyn->d_un.d_ptr = findSection(".gnu.version_r").sh_addr;
             else if (d_tag == DT_VERSYM)
                 dyn->d_un.d_ptr = findSection(".gnu.version").sh_addr;
+            else if (d_tag == DT_MIPS_RLD_MAP_REL) {
+                /* the MIPS_RLD_MAP_REL tag stores the offset to the debug
+                   pointer, relative to the address of the tag */
+                debug("Updating DT_MIPS_RLD_MAP_REL");
```
Mic92, line 1147: nitpick. we don't do any debug printing for other sections:

```suggestion
```

src/patchelf.cc
```diff
@@ -1129,6 +1141,14 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
                 dyn->d_un.d_ptr = findSection(".gnu.version_r").sh_addr;
             else if (d_tag == DT_VERSYM)
                 dyn->d_un.d_ptr = findSection(".gnu.version").sh_addr;
+            else if (d_tag == DT_MIPS_RLD_MAP_REL) {
```
Mic92, line 1144: Could this section be stripped just like `DT_JMPREL` or `DT_REL`?

Mic92: What I am asking is, is it possible to that someone has binaries where this section was stripped.
I fixed a similar issue here: https://github.com/NixOS/patchelf/pull/296

Mic92:
I also gained now merge rights for this repo.

src/patchelf.cc
iv-m: Ok, I'm going to drop this.

src/patchelf.cc
iv-m: @Mic92, thank you for the review.

I understood what you're asking, and I need to give it some thought and run a couple of experiments.

I used to believe that elf with `DT_MIPS_RLD_MAP_REL` in dynamic section but without `.rld_map` is simply broken, but maybe patchelf needs to handle such files. If that's the case, I'll update the code here and add a test for that.

iv-m: Well, elf file with `DT_MIPS_RLD_MAP_REL` in dynamic section but without `.rld_map` is indeed broken: dynamic loader (I've checked glibc 2.32) just writes the debug pointer to the address it calculates from  `DT_MIPS_RLD_MAP_REL` entry. I guess no tool produces such binaries.

But when I just removed `.rtld_map` section with `objcopy`, the resulting binaries mostly worked -- out of sheer luck probably, but still most of them did not crash.

So I rewrote the code to handle such case. Removing `DT_MIPS_RLD_MAP_REL` might have been a  cleaner solution, but `patchelf` does not remove any dynamic section entries yet and  I don't think supporting binaries that were broken in the fist place worth the burden of starting to do so.

And @Mic92, thank you for pointing this case out.

Mic92:
Thanks!

PR #198, opened 2020-04-08
Update doc to describe multiple filename args allowed

pespin:
Commit 936bae418b77ee9e06a93cd3cd444f4204446973 added support to handle
several files at once, but forgot to update the man and help output to
describe it.

Fixes: 936bae418b77ee9e06a93cd3cd444f4204446973


edolstra:
Thanks!

PR #201, opened 2020-06-01
Added --output flag

dstahlke:
This flag allows the output to be different from the input.  This can be used during `make install` to avoid an unnecessary cp command.  That can save a lot of disk I/O for large projects.


edolstra:
Thanks!

Issue #170, opened 2019-05-17
patchelf 0.10 corrupts some shared libraries


sjackman:
After using `patchelf` 0.10 to patch a shared library,  ldd reports ``Inconsistency detected by ld.so: dl-version.c: 224: _dl_check_map_versions: Assertion `needed != NULL' failed!``.

Steps to reproduce:
```console
$ curl -L https://linuxbrew.bintray.com/bottles/glib-2.60.2.x86_64_linux.bottle.tar.gz | tar xz
$ patchelf --print-needed glib/2.60.2/lib/libglib-2.0.so.0.6000.2
libpcre.so.1
libpthread.so.0
libc.so.6
$ patchelf --print-rpath glib/2.60.2/lib/libglib-2.0.so.0.6000.2
@@HOMEBREW_PREFIX@@/lib
$ ldd glib/2.60.2/lib/libglib-2.0.so.0.6000.2
	linux-vdso.so.1 (0x00007ffce67eb000)
	libpcre.so.1 => not found
	libpthread.so.0 => /gsc/btl/linuxbrew/Cellar/glibc/2.23/lib/libpthread.so.0 (0x00007f36cfc68000)
	libc.so.6 => /gsc/btl/linuxbrew/Cellar/glibc/2.23/lib/libc.so.6 (0x00007f36cf8cc000)
	/gsc/btl/linuxbrew/Cellar/glibc/2.23/lib64/ld-linux-x86-64.so.2 (0x00007f36d01be000)
$ chmod +w glib/2.60.2/lib/libglib-2.0.so.0.6000.2
$ patchelf --set-rpath /home/linuxbrew/.linuxbrew/lib glib/2.60.2/lib/libglib-2.0.so.0.6000.2
$ patchelf --print-needed glib/2.60.2/lib/libglib-2.0.so.0.6000.2
$ patchelf --print-rpath glib/2.60.2/lib/libglib-2.0.so.0.6000.2

$ ldd glib/2.60.2/lib/libglib-2.0.so.0.6000.2
Inconsistency detected by ld.so: dl-version.c: 224: _dl_check_map_versions: Assertion `needed != NULL' failed!
```

sjackman:
In case it's relevant, this shared library was previously patched using `patchelf` 0.9 to change the RPATH, and now it's being patched using `patchelf` 0.10 to change the RPATH again.

iMichka:
[...]

Issue #192, opened 2020-02-11
binary fails to load (_dl_check_map_versions) after patchelf --set-rpath 


pespin:
Original ticket tacking the issue: https://osmocom.org/issues/4389

So, in summary:
I was using patchelf 0.9-1+b1 from debian9 repos, but run into issue #47 with binary "srsue" from srsLTE.git [1]:
```
$ patchelf --set-rpath /osmo-gsm-tester-srsue/srslte/lib /osmo-gsm-tester-srsue/srslte/bin/srsue
warning: working around a Linux kernel bug by creating a hole of 124461056 bytes in ‘/osmo-gsm-tester-srsue/srslte/bin/srsue’
maximum file size exceeded
```

 so I then went into building/installing a new version of patchelf. I tried with following patchelf versions:
* 0.9+52 (debian10 patchelf_0.9+52.20180509-1_amd64.deb)
* 0.10 (last release Download tarball and bult manually)
* master (from git, 2ba64817ec6f3b714503ea6e6aa8439505bb7393)

All of them fix the issue #47 on binary "srsue". However, all those new versions introduce a new bug with another binary generated by srsLTE.git [1] called "srspec". Patching itself seems to be working fine (patchelf doesn't exit with !=0), but afterwards the loader is unable to load the binary:
```
# ldd /osmo-gsm-tester-srsepc/srslte/bin/srsepc
Inconsistency detected by ld.so: dl-version.c: 224: _dl_check_map_versions: Assertion `needed != NULL' failed!
```
This issue doesn't show up with the original patchelf 0.9-1+b1 from debian9 I was using, so it's a regression added at some point after that and yet not fixed last release nor in currentmaster.

[1] https://github.com/srsLTE/srsLTE

pespin:
I attach a diff here of readelf launched against the working binary (patchelf 0.9-1+b1) and the broken binary (master 2ba64817ec6f3b714503ea6e6aa8439505bb7393).
[diff.log](https://github.com/NixOS/patchelf/files/4185602/diff.log)

pespin:
According to git bisect, this is the commit introducing the regression: c4deb5e9e1ce9c98a48e0d5bb37d87739b8cfee4

```
$ git bisect start
$ git bisect bad master
$ git bisect good 0.9
[...]

PR #202, opened 2020-06-03
Fix shared library corruption when rerunning patchelf

rpurdie:
When running patchelf on some existing patchelf'd binaries to change to longer
RPATHS, ldd would report the binaries as invalid. The output of objdump -x on
those libraryies should show the top of the .dynamic section is getting trashed,
something like:

0x600000001 0x0000000000429000
0x335000 0x0000000000335000
0xc740 0x000000000000c740
0x1000 0x0000000000009098
SONAME libglib-2.0.so.0

(which should be RPATH and DT_NEEDED entries)

This was tracked down to the code which injects the PT_LOAD section.

The issue is that if the program headers were previously relocated to the end
of the file which was how patchelf operated previously, the relocation code
wouldn't work properly on a second run as it now assumes they're located after
the elf header. This change forces them back to immediately follow the elf
header which is where the code has made space for them.

Should fix https://github.com/NixOS/patchelf/issues/170
and https://github.com/NixOS/patchelf/issues/192

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -754,6 +754,7 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
     }
 
     /* Add a segment that maps the replaced sections into memory. */
+    wri(hdr->e_phoff, sizeof(Elf_Ehdr));
     phdrs.resize(rdi(hdr->e_phnum) + 1);
     wri(hdr->e_phnum, rdi(hdr->e_phnum) + 1);
     Elf_Phdr & phdr = phdrs[rdi(hdr->e_phnum) - 1];


domenkozar:
@rpurdie any chance to add a test case in tests directory?

rpurdie:
The easiest test case would be to check in the shared lib from the testcase in #170, run patchelf against it as described and check ldd finds the binary is still valid. Not sure if you can add that lib though and creating such a lib from sources would be a much more involved exercise.

domenkozar:
I'm going to merge this and bump patchelfUnstable in nixpkgs. Then we can build all packages and check for possible regressions.

sjackman:
🎉 ❤️ Thank you very much for this fix, Richard!

sjackman:
@domenkozar Is there a stable release of `patchelf` planned that includes this fix?

domenkozar:
Yes, you can follow https://github.com/NixOS/nixpkgs/issues/69213 as we rebuild all nix packages to verify there are no regressions this time. Once that's done we'll make a release.

PR #203, opened 2020-06-03
fix README -> README.me in some more places

priegger:
- this fixes `nix-build release.nix`
- rpm build also works (tested `patchelf --verison` on fedora 27 in docker)


domenkozar:
Thanks!

PR #205, opened 2020-06-03
run CI with musl as well

domenkozar:
Depends on https://github.com/NixOS/nixpkgs/pull/89425


domenkozar:
Will be based on nixpkgs-unstable once musl packages hit hydra via https://hydra.nixos.org/jobset/nixos/trunk-combined#tabs-evaluations

domenkozar:
I ran out of time to debug this, it works for me locally. Would be great if someone else can take a look.

PR #206, opened 2020-06-04
allow overriding systems

priegger:
This allows skipping the aarch64 build on x86-only hydra instances.

Note: I copied the systems list with all commented systems.

In the nix release.nix the parameter is called "systems", in the nixpkgs nixos/release.nix it is called "supportedSystems". I went with "systems" but can also change it.


domenkozar:
I've done this in https://github.com/NixOS/patchelf/pull/205/files#diff-6ab84d77bf6d73598ebba867195ce313R5, which I hope to merge in a few days

priegger:
Will be fixed in #205.

PR #218, opened 2020-06-20
Better support relocating NOTE sections/segments

### LINKED ISSUE #217: patchelf doesn't update PT_NOTE segment when relocating SHT_NOTE sections
**Describe the bug**

`SHT_NOTE` sections (e.g. `.note.gnu.build-id`) can sometimes be aliased by a `PT_NOTE` segment. When this is the case, patchelf might move the `SHT_NOTE` section without also updating the `PT_NOTE` segment. The `PT_NOTE` segment ends up pointing to the middle of random stuff that might have been overwritten.

This is as far as I know the root cause of https://github.com/NixOS/nixpkgs/issues/91145 breaking Mesa in NixOS with patchelf 0.11.

**Steps To Reproduce**

```
$ echo 'int foo(int x) { return x; }' > test.c
$ gcc -s -shared -Wl,-build-id -o test.so test.c
$ readelf -n test.so
Displaying notes found in: .note.gnu.build-id
  Owner                 Data size	Description
  GNU                  0x00000014	NT_GNU_BUILD_ID (unique build ID bitstring)
    Build ID: 767c996573519ca3392910e472102d68d87342e7

$ patchelf --set-rpath $(perl -e 'print "A"x1000') test.so
$ readelf -n test.so
Displaying notes found in: .note.gnu.build-id
  Owner                 Data size	Description
readelf: Warning: note with invalid namesz and/or descsz found at offset 0x0
readelf: Warning:  type: 0x3, namesize: 0x00000004, descsize: 0x00000014, alignment: 8
```

**Expected behavior**

Not completely sure yet. `PT_NOTE` should be updated, but that means we might need to ensure we keep multiple note sections adjacent if the `PT_NOTE` contains multiple `SHT_NOTE` sections.

**`patchelf --version` output**

master

delroth:
Fixes #217 

SHT_NOTE sections can be mapped in memory by PT_NOTE segments. When
rewriting an SHT_NOTE, attempt to also rewrite a matching segment if it
exists.

Note that an ELF can contain multiple SHT_NOTE sections, and a given
PT_NOTE segment can theoretically map multiple contiguous sections. This
is not currently supported, but we fail loudly when encountering it.

Also fix (or more like hack around) alignment issues with note sections.
Keeping the original alignment value when possible is important because
it determines how the data within the section needs to be parsed.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -187,6 +187,8 @@ private:
 
     void rewriteSectionsExecutable();
 
+    void normalizeNoteSegments();
+
 public:
 
     void rewriteSections();
@@ -648,6 +650,7 @@ void ElfFile<ElfFileParamNames>::writeReplacedSections(Elf_Off & curOff,
     for (auto & i : replacedSections) {
         std::string sectionName = i.first;
         auto & shdr = findSection(sectionName);
+        Elf_Shdr orig_shdr = shdr;
         debug("rewriting section '%s' from offset 0x%x (size %d) to offset 0x%x (size %d)\n",
             sectionName.c_str(), rdi(shdr.sh_offset), rdi(shdr.sh_size), curOff, i.second.size());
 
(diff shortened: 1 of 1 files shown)


delroth:
Hmm, looks like this won't be enough. The situation with multiple SHT_NOTE covered by a single PT_NOTE is common enough that apparently we hit it even in the test suite :-/

```
rewriting section '.note.ABI-tag' from offset 0x121c (size 32) to offset 0x578 (size 32)
patchelf: unsupported overlap of SHT_NOTE and PT_NOTE
FAIL no-rpath-amd64.sh (exit status: 1)
```

And indeed there is a single `PT_NOTE` covering both `.note.ABI-tag` and `.note.gnu.build-id`.

Not completely sure what the best way forward is -- maybe some kind of normalization step where we break up the `PT_NOTE` into multiple segments. I'll give that a shot unless someone comes up with a better idea.

delroth:
I implemented the normalization. Tests now pass, and as far as I can tell in my limited testing everything seems to be working fine.

flokli:
@domenkozar, can you take a look at this? This should fix https://github.com/NixOS/nixpkgs/issues/91145.

domenkozar:
I'm not familiar with the codebase, maybe @edolstra can take a look.

Mic92:
Does any of the tests exercise this segment?

Mic92:
I also added a regression test in https://github.com/NixOS/patchelf/pull/225

Mic92:
Feel free to cherry pick.

edolstra:
Thanks, I've merged #225 which includes this one!

PR #219, opened 2020-06-23
CI for aarch64

domenkozar:


.travis.yml
```diff
@@ -0,0 +1,4 @@
+arch:
+  - arm64
```
Mic92, line 2: Is this self-hosted?

domenkozar: Nope, travis supports aarch64

PR #221, opened 2020-06-29
more output, default to print all

vanhauser-thc:
if several -print-... options are given it is unclear which output refers to what as everything is just printed as `printf("<item>\n");`. This puts an identified before the entry to clarify this.

If a lib for removing was not found, no error/warning was given, this adds a success/failure message.

Also if no parameters are given it defaults to "print all", which is less to type and more convenient than a no-op as previously.

I did these changes just for me, if you like them accept, if not its fine  :)

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1083,7 +1083,7 @@ void ElfFile<ElfFileParamNames>::modifySoname(sonameMode op, const std::string &
             if (std::string(soname ? soname : "") == "")
                 debug("DT_SONAME is empty\n");
             else
-                printf("%s\n", soname);
+                printf("DT_SONAME: %s\n", soname);
         } else {
             debug("no DT_SONAME found\n");
         }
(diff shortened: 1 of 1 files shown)


src/patchelf.cc
```diff
@@ -1585,7 +1591,7 @@ template<class ElfFile>
 static void patchElf2(ElfFile && elfFile, const FileContents & fileContents, std::string fileName)
 {
     if (printInterpreter)
-        printf("%s\n", elfFile.getInterpreter().c_str());
+        printf("INTERPRETER: %s\n", elfFile.getInterpreter().c_str());
```
Mic92, line 1594: Sorry. We cannot merge this as this would break our use in nixpkgs, which expects no prefix in many places.

Mic92:
In particular it breaks the following usecase:

```console
$ foo=$(patchelf --print-interpreter /run/current-system/sw/bin/ls)
$ echo $foo
/nix/store/cvr0kjg2q7z2wwhjblx6c73rv422k8cm-glibc-2.33-47/lib/ld-linux-x86-64.so.2
```

PR #225, opened 2020-07-31
Better support relocating NOTE sections/segments + tests!

Mic92:


haampie:
This PR generates executuables where the `.dynstr` section is not fully contained in a segment, but overlaps two segments. `gdb` complains about this, and on a particular filesystem I'm using, it seems to cause the binary to be broken (it errors with mmap/mprotect errors or segfaults)

Mic92:
The regression hopefully fixed now.

PR #227, opened 2020-08-11
use stable_sort to maintain relative order for equal elements in sections

rmNULL:
replaces the usage of sort with stable_sort as ordering for equal elements is undefined for std::sort.


sjackman:
Thanks for merging, Eelco!

PR #234, opened 2020-08-31
replace language:nix

abathur:
Draft PR to make CI run, per request from @Mic92 :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,7 @@ arch:
   - arm64
 version: ~> 1.0
 import: nix-community/nix-travis-ci:nix.yml@main
+env:
+  global:
+    - SKIP_ADDING_NIXPKGS_CHANNEL: true
 script: nix-build release.nix -A build.aarch64-linux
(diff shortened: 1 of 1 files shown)


Mic92:
The background here is that we plan to move travis ci support for nix in a separate repository.
Right now it is supported by travis upstream however this integration is not well maintained and
changes take too long to get in. Having versioned nix installer allows smoother upgrades similar to how the install-nix github action works. See also https://discourse.nixos.org/t/fate-of-travis-ci-nix-language-support/7907/12

Mic92:
Any reason to still mark it as a draft?

abathur:
@Mic92 No functional reason; I just moved this whole effort to the backburner for a few days to try to get resholve~~d~~ into nixpkgs before branch-off, so I haven't done any more here aside from address your comments. (I've got tabs with the rest of the nix-community repos that use travis-ci, for example, to PR when I have a little more time...)

Mic92:
cc @domenkozar for review. We are also using this in NUR for quite some time.

Issue #158, opened 2018-09-19
--replace-needed calls corrupt these executables


sztomi:
**patchelf version:** Built from 27ffe8a

In our build process, we try to remove all version numbers from the shared objects names of our various dependencies. We use patchelf to fix the referenced sonames everywhere. This seems to work well for the most part, but we noticed one particular example where it corrupts the executable.

In the first case, there were no version numbers in the sonames, so the rewriting is redundant. Download the binary [here](https://drive.google.com/file/d/15kT55eNLem07Mn6NYcV4FcE7bxRN-yuW/view?usp=sharing)

```
$ patchelf --print-needed ./ffmpeg
libavfilter.so
libavformat.so
libavcodec.so
libswresample.so
libswscale.so
libavutil.so
libm.so
libc.so

$ patchelf --replace-needed 'libavfilter.so' 'libavfilter.so' --replace-needed 'libavformat.so' 'libavformat.so' --replace-needed 'libavcodec.so' 'libavcodec.so' --replace-needed 'libswresample.so' 'libswresample.so' --replace-needed 'libswscale.so' 'libswscale.so' --replace-needed 'libavutil.so' 'libavutil.so' ./ffmpeg

$ patchelf --print-needed ./ffmpeg
libavutil.so
bavcodec.so
avformat.so
libswresample.so
bswscale.so
bavfilter.so
libm.so
libc.so
```

Notice how `libavcodec.so` became `bavcodec.so` etc. 

In the second case, there are numbers to strip. Download the binary [here](https://drive.google.com/file/d/1iH-wUnDSsta0yiLfTdpniZtha-2oPpHu/view?usp=sharing)

```
$ patchelf --print-needed ./ffmpeg2
libavfilter.so.7
libavformat.so.58
libavcodec.so.58
libswresample.so.3
libswscale.so.5
libavutil.so.56
libm.so.6
libpthread.so.0
libc.so.6

$ patchelf --replace-needed 'libavfilter.so.7' 'libavfilter.so' --replace-needed 'libavformat.so.58' 'libavformat.so' --replace-needed 'libavcodec.so.58' 'libavcodec.so' --replace-needed 'libswresample.so.3' 'libswresample.so' --replace-needed 'libavutil.so.56' 'libavutil.so' --replace-needed 'libm.so.6' 'libm.so' --replace-needed 'libpthread.so.0' 'libpthread.so' --replace-needed 'libc.so.6' 'libc.so' ./ffmpeg2

[...]

PR #237, opened 2020-09-17
avoid string table duplicates and DT_NEEDED reference corruption when using --replace-needed multiple times

DerDakon:
When it happens that the .gnu.version_r stores the strings in .dynstr it can come to corruption of the library names written into DT_NEEDED (#158):
    
- the library names in DT_NEEDED are replaced, new entries are written to the end of .dynstr
- the version library names are replaced, and written to the end of the string section.
    
If the section for the version strings is also ".dynstr", the previous modifications were _not_ taken into account and things were written from the old end of .dynstr again. The order in which these strings were written is not the same as the previous replacement, so things would end up with the same size, but different offsets. The .gnu.version_r table is correct, the file contents are fine, but the offsets in the DT_NEEDED entries are wrong. Since they are printed as 0-terminated strings the first one replaced will always be shown correct, which also is the case if the argument is only used once as the string is replaced with itself afterwards.

Now things are correct, but the strings are added to .dynstr twice. Use an unordered map to cache them.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1551,6 +1551,10 @@ void ElfFile<ElfFileParamNames>::replaceNeeded(const std::map<std::string, std::
         debug("found .gnu.version_r with %i entries, strings in %s\n", verNeedNum, versionRStringsSName.c_str());
 
         unsigned int verStrAddedBytes = 0;
+        // It may be that it is .dynstr again, in which case we must take the already
+        // added bytes into account.
+        if (versionRStringsSName == ".dynstr")
+            verStrAddedBytes += dynStrAddedBytes;
 
         auto need = (Elf_Verneed *)(contents + rdi(shdrVersionR.sh_offset));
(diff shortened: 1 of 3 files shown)


src/patchelf.cc
```diff
@@ -1544,6 +1556,13 @@ void ElfFile<ElfFileParamNames>::replaceNeeded(const std::map<std::string, std::
         debug("found .gnu.version_r with %i entries, strings in %s\n", verNeedNum, versionRStringsSName.c_str());
 
         unsigned int verStrAddedBytes = 0;
+        // it may be that is is .dynstr again, in which case we must take the already
```
Mic92, line 1559:
```suggestion
        // it may be that is  .dynstr again, in which case we must take the already
```

DerDakon: That should have been "it is", thanks for spotting.

Mic92:
Thanks! I rebased and merged this.

Issue #239, opened 2020-09-17
Library corruption when setting RPATH


pablogsal:
**Describe the bug**

Some shared libraries are corrupted when `patchelf` sets the `RPATH` to some new value.

**Steps To Reproduce**

Consider this shared library:

https://www.dropbox.com/s/n47wla3tgw4ny56/libfreebl3.so?dl=0

If you create a very simple executable that links against it, everything works:

```
echo "int main(){return 0;}" > lel.c
gcc lel.c -L . -lfreebl3
./a.out
```

but now, changing the RPATH with patchelf (master) makes the program segfault:

```
./src/patchelf --set-rpath $ORIGIN --force-rpath libfreebl3.so
❯ ./a.out
[1]    19792 segmentation fault (core dumped)  ./a.out
```

Apparently, the segfault happens when the linker is resolving relocations:

```
LD_DEBUG=all ./a.out

...
     19907:     file=libdl.so.2 [0];  generating link map
     19907:       dynamic: 0x00007fe6ad83ecf0  base: 0x00007fe6ad83a000   size: 0x0000000000005090
     19907:         entry: 0x00007fe6ad83b210  phdr: 0x00007fe6ad83a040  phnum:                 11
     19907:
     19907:     checking for version `GLIBC_2.2.5' in file /usr/lib/libc.so.6 [0] required by file ./a.out [0]
     19907:     checking for version `GLIBC_2.2.5' in file /usr/lib/libdl.so.2 [0] required by file libfreebl3.so [0]
     19907:     checking for version `GLIBC_2.3' in file /usr/lib/libc.so.6 [0] required by file libfreebl3.so [0]
     19907:     checking for version `GLIBC_2.2.5' in file /usr/lib/libc.so.6 [0] required by file libfreebl3.so [0]
     19907:     checking for version `GLIBC_2.2.5' in file /lib64/ld-linux-x86-64.so.2 [0] required by file /usr/lib/libc.so.6 [0]
     19907:     checking for version `GLIBC_2.3' in file /lib64/ld-linux-x86-64.so.2 [0] required by file /usr/lib/libc.so.6 [0]
     19907:     checking for version `GLIBC_PRIVATE' in file /lib64/ld-linux-x86-64.so.2 [0] required by file /usr/lib/libc.so.6 [0]
     19907:     checking for version `GLIBC_PRIVATE' in file /lib64/ld-linux-x86-64.so.2 [0] required by file /usr/lib/libdl.so.2 [0]
[...]

PR #242, opened 2020-09-22
Use sh_offset instead of sh_addr when checking already replaced libs

pablogsal:
When checking for already replaced libs, the check against the size must
be done using the section header offset, not the section file address.
This was not crashing in many situations because normally sh_address and
sh_offset have the same value but these two may differ and using the
sh_address value instead can cause library corruption in these
situations.

Closes: #239


pablogsal:
Note: I am not including a test because unfortunately, producing libraries with this setup requires some linker scripts as it needs to modify the program header address. and that will complicate considerably the test suite in order to make it portable. Given that this code path is always hit, we still have certainty that works as expected in previous cases.

The only way that I can think of including a test is vendoring some of the libraries we know that have problems.

pablogsal:
CC: @edolstra @domenkozar

PR #246, opened 2020-10-05
Added option --add-rpath

xavierabellan:
New option --add-rpath makes it possible to just add a new entry to the rpath list preserving the existing ones in one operation. This avoids having to run  patchelf --print-rpath, save the current value in a variable, append the new path to the variable and then patchelf --set-rpath.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1939,6 +1939,9 @@ int mainWrapped(int argc, char * * argv)
     if (!outputFileName.empty() && fileNames.size() != 1)
         error("--output option only allowed with single input file");
 
+    if (setRPath && addRPath)
+        error("--set-rpath option not allowed with --add-rpath");
+
     patchElf();
 
     return 0;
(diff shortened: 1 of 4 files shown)


src/patchelf.cc
```diff
@@ -1748,6 +1753,8 @@ static void patchElf2(ElfFile && elfFile, const FileContents & fileContents, std
         elfFile.modifyRPath(elfFile.rpRemove, {}, "");
     else if (setRPath)
         elfFile.modifyRPath(elfFile.rpSet, {}, newRPath);
+    else if (addRPath)
+        elfFile.modifyRPath(elfFile.rpAdd, {}, newRPath);
```
Mic92, line 1757: Should there be a conflict between add rpath and set rpath?

xavierabellan: Thanks for looking into this. I've just added a conflict when parsing arguments so you can only use one of them.

Mic92 (approved):
The option makes the tool easier to use and its test should cover all cases.

PR #251, opened 2020-11-15
Fuzzing fixes

blitz:
I've been running patchelf against the [afl](https://lcamtuf.coredump.cx/afl/) fuzzer, because we've seen `patchelf` crash in our production environment. This MR contains test cases and fixes for the first round of issues.

I've been operating with this assumption:
- `patchelf` should never trigger assertions or crash for malformed ELF files.

I've structured this MR in this way:
- commit to introduce test case for issue X
- commit to fix issue X
- commit to introduce test case for issue Y
- commit to fix issue Y
- ...

There are a few trivial fixes that come from code reading that currently have no corresponding test case, because the fuzzer couldn't construct one. If this is an issue, I can probably manually construct an offending file.

@tfc @domenkozar


blitz:
With this patch of fixes, AFL now runs for 3h without finding anything. This is a big step up from before. :)

edolstra:
Thanks!

Issue #88, opened 2016-04-03
hardcoded maxSize?


efferre79:
During a --set-rpath operation on a vmware binary (vmware-vmx-debug) I get a "maximum file size exceeded" error. After looking at the code I have seen that the maxSize is calculated adding 32MB to the file size. Is there any reason for this limit?

aitzkora:
I observed also this limitation. I have to patch some file up to 1.4G, I think I will used a patched version of patchelf. Its name couild be patchpatchelf :smile:

domenkozar:
Fixed in #256

Issue #62, opened 2015-08-17
Unable to run patchelf on some Ubuntu binaries


n3rdopolis:
Namely on the dynamically linked qemu-user (non-static) binaries. I'm not sure how Ubuntu compiles them

With the newest version of patchelf compiled:
I try
./patchelf --set-interpreter /path/to/ld-linux.so.2 /path/to/qemu-arm
and get
warning: working around a Linux kernel bug by creating a hole of 33796096 bytes in ‘/path/to/qemu-arm’
maximum file size exceeded

I get the same error with --set-rpath as well on the same binary

n3rdopolis:
I found that there is debug flag

patching ELF file `/tmp/qemu_user_server/arm/qemu-user'
Kernel page size is 4096 bytes
new rpath is `/tmp/qemu_user_server/arm/'
rpath is too long, resizing...
DT_NULL index is 33
replacing section `.dynamic' with size 312
replacing section `.dynstr' with size 16703
this is a dynamic library
last page is 0x21d9000
needed space is 17368
warning: working around a Linux kernel bug by creating a hole of 33796096 bytes in ‘/tmp/qemu_user_server/arm/qemu-user’
maximum file size exceeded

DoomHammer:
Had exactly the same issue, but worked around (this work around) by increasing the `maxSize` to `64 * 1024 * 1024` in [patchelf.cc:245](https://github.com/NixOS/patchelf/blob/master/src/patchelf.cc#L245) which was just about right.

The questions that arise are:
1. @edolstra why is `maxSize` set to 32M (previously 64M)? Why not 512M or even 42G?
2. @edolstra again: you mention Linux kernel bug in your commit message but don't exactly reference it. Has it been, perchance, already fixed and thus rendered the workaround unnecessary (for newer kernels)?

n3rdopolis:
That workaround works, but it creates a huge 34 MB executable...

I should post some readelf info
readelf -d -e /usr/bin/qemu-i386 
ELF Header:
  Magic:   7f 45 4c 46 01 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF32
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Intel 80386
  Version:                           0x1
  Entry point address:               0x30f9c
  Start of program headers:          52 (bytes into file)
[...]

PR #256, opened 2020-12-05
Remove limits on output file size.

andrewla:
> The use of self-pointers into a vector<> meant that the vector could not be relocated, which meant that any growth in the file had to fit within the initial reservation.

> Now, instead of a pointer to a header and a pointer to the contents, those values are computed dynamically from the underlying vector().

I made an attempt at writing a test that would provide coverage, but couldn't make it work. If you have suggestions for writing a test to test this functionality, please let me know and I'll give it a shot.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ Makefile
 /tests/main-scoped
 /tests/libbig-dynstr.debug
 /tests/contiguous-note-sections
+/tests/simple-pie
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -94,9 +94,6 @@ public:
 
 private:
 
-    unsigned char * contents;
-
-    Elf_Ehdr * hdr;
     std::vector<Elf_Phdr> phdrs;
     std::vector<Elf_Shdr> shdrs;
 
@@ -233,6 +230,14 @@ private:
         t = rdi((I) i);
         return i;
     }
+
+    Elf_Ehdr *hdr() {
+      return (Elf_Ehdr *)fileContents->data();
+    }
+
+    const Elf_Ehdr *hdr() const {
+      return (const Elf_Ehdr *)fileContents->data();
+    }
 };
 
 
@@ -305,13 +310,6 @@ __attribute__((noreturn)) static void error(const std::string & msg)
     throw std::runtime_error(msg);
 }
 
-static void growFile(const FileContents & contents, size_t newSize)
-{
-    if (newSize > contents->capacity()) error("maximum file size exceeded");
-    if (newSize <= contents->size()) return;
-    contents->resize(newSize, 0);
-}
-
 static FileContents readFile(const std::string & fileName,
     size_t cutOff = std::numeric_limits<size_t>::max())
 {
@@ -324,9 +322,7 @@ static FileContents readFile(const std::string & fileName,
 
     size_t size = std::min(cutOff, static_cast<size_t>(st.st_size));
 
-    FileContents contents = std::make_shared<std::vector<unsigned char>>();
-    contents->reserve(size + 32 * 1024 * 1024);
-    contents->resize(size, 0);
+    FileContents contents = std::make_shared<std::vector<unsigned char>>(size);
 
     int fd = open(fileName.c_str(), O_RDONLY);
     if (fd == -1) throw SysError(fmt("opening '", fileName, "'"));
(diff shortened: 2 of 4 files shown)


andrewla:
Fixes #88 
Fixes #62

domenkozar:
@andrewla I know this is probably not helpful, but PRs with tests have a higher chance of getting merged. 

I do not have ideas how to make a dummy big output.

andrewla:
I'm going to put this on hold. As a stress tests, I used this as an overlay for patchelf in nix, I'm getting errors trying to build gcc.

Not clear if this is patchelf related or some configuration problem, but for the moment I don't consider this safe to merge.

dstahlke:
FWIW, this patch seems to work for me.  I tried this patch, and also tried just boosting the buffer size to 256MB.  The output of each method was identical.  Unfortunately I am not able to share my test case, because of IP reasons.  I haven't reviewed the changes to make sure there are no lingering pointers, but at least it passes valgrind.

andrewla:
I'm all but positive that the errors encountered are related to how nix does bootstrapping of the compiler, rather than related to this change specifically.

Mic92:
Hey. I joined the maintainer team of patchelf recently if you could resolve the merge conflicts I give this a test.

andrewla:
I applied the latest changes; I get a strange test failure locally because it can't find the "strip" utility but that appears to be some autoconf issue that is probably local to my machine; running the strip manually (to produce `libbig-dynstr.debug` makes the test pass).

Mic92:
Does `make STRIP=strip` fixes that? Might be because of https://github.com/NixOS/patchelf/blob/c62988a2a89a282e31d0e204cc2a3710da912f04/tests/Makefile.am#L3

Mic92:
So one way to test this is to link in a large binary blob (can be created with `dd`): https://balau82.wordpress.com/2012/02/19/linking-a-binary-blob-with-gcc/ and than create a case where this section needs to be replaced.
I am currently working on a deadline for a different project so I might look later at this unless you come up with a test first.

trofi:
I think the problem is not in original binary size, but in limited growth that `patchelf` allows for. We can pick a tiny hello world and try to add 32MB of data to it. Say, 32MB of NEEDED libraries:

```
$ nix-shell -p patchelf

# very clumsy way to achieve many long "--add-needed AAAAAAAAA...." strings.
[nix-shell:/tmp]$ echo 'int main(){}' | gcc -x c - -o a -fPIC -pie; ./a; while patchelf --add-needed $(perl -e 'print "a" x 100000;') --add-needed $(perl -e 'print "b" x 100000;') --add-needed $(perl -e 'print "c" x 100000;') --add-needed $(perl -e 'print "d" x 100000;') --add-needed $(perl -e 'print "e" x 100000;') --add-needed $(perl -e 'print "f" x 100000') --add-needed $(perl -e 'print "g" x 100000') --add-needed $(perl -e 'print "h" x 100000') --add-needed $(perl -e 'print "i" x 100000') --add-needed $(perl -e 'print "j" x 100000') --add-needed $(perl -e 'print "k" x 100000') --add-needed $(perl -e 'print "l" x 100000') --add-needed $(perl -e 'print "m" x 100000') --add-needed $(perl -e 'print "m" x 100000') --add-needed $(perl -e 'print "o" x 100000') --add-needed $(perl -e 'print "p" x 100000') --add-needed $(perl -e 'print "q" x 100000') --add-needed $(perl -e 'print "r" x 100000') --add-needed $(perl -e 'print "s" x 100000') --add-needed $(perl -e 'print "t" x 100000') a; do echo again; done
again
again
again
again
again
again
again
again
again
again
again
again
again
again
again
again
again
patchelf: maximum file size exceeded
```

samuela:
Ran into this issue while attempting to package tensorflow 2.6.0:
```
Rewriting #!/nix/store/1g54cwl6m78rgrqsnn1pp29dbvldpvsd-python3-3.9.6/bin/python3.9 to #!/nix/store/1g54cwl6m78rgrqsnn1pp29dbvldpvsd-python3-3.9.6
wrapping `/nix/store/hcghxmsg8smr0fmdpq9nr2pfifgcz586-python3.9-tensorflow-gpu-2.6.0/bin/tflite_convert'...
patchelf: maximum file size exceeded
builder for '/nix/store/gkp1bdbkya9p5zgxrbbp70qks31zmsz3-python3.9-tensorflow-gpu-2.6.0.drv' failed with exit code 1
```

samuela:
Oddly enough, I've not been able to reproduce the failure outside of the nix build environment. I've been trying to provide a test case for this, but patchelf isn't failing outside of nix-build. In my `postFixup`:
```
for $lib in ...
      echo "processing patchelf"
      cp "$lib" "$lib.orig"
      echo "$lib"
      echo $(patchelf --print-rpath "$lib")
      patchelf --set-rpath "${cudatoolkit}/lib:${cudatoolkit.lib}/lib:${cudnn}/lib:${nccl}/lib:$(patchelf --print-rpath "$lib")" "$lib"
done
```
And I get output:
```
processing patchelf
/nix/store/i5ywd496x0ppp1lifkfwncsmixdsvm9s-python3.9-tensorflow-gpu-2.6.0/lib/python3.9/site-packages/tensorflow/python/_pywrap_toco_api.so
/run/opengl-driver/lib:$ORIGIN/../../_solib_local/_U_S_Stensorflow_Spython_C_Upywrap_Utoco_Uapi.so___Utensorflow:$ORIGIN/../../_solib_local/_U_S_Stensorflow_Spython_C_Upywrap_Utensorflow_Uinternal_Ulinux___Utensorflow_Spython:$ORIGIN/:$ORIGIN/..:/nix/store/g3h82za6g1j3gwz1br0spzpx1f20hc3c-tensorflow-gpu-2.6.0/lib:/nix/store/9bh3986bpragfjmr32gay8p95k91q4gy-glibc-2.33-47/lib:/nix/store/ql152marawkw62kvkjskndxccffdfns1-gcc-9.3.0-lib/lib
patchelf: maximum file size exceeded
exitHandler
_callImplicitHook 0 failureHook
builder for '/nix/store/1cvkfzsnpiz31qp0wpimqgiqhgj95xf3-python3.9-tensorflow-gpu-2.6.0.drv' failed with exit code 1
```
(You can find the full code to reproduce in https://github.com/NixOS/nixpkgs/pull/137069, `nix-build -A tensorflowWithCuda`.)

But when I try to test patchelf outside of nix-build:
```
lib="/nix/store/i5ywd496x0ppp1lifkfwncsmixdsvm9s-python3.9-tensorflow-gpu-2.6.0/lib/python3.9/site-packages/tensorflow/python/_pywrap_toco_api.so.orig"
cp "$lib" /tmp/deleteme.so
patchelf --set-rpath "/nix/store/7z8w0varw65bfvgqb82lm4isggvfap23-cudatoolkit-11.0.3/lib:/nix/store/94nh9ffcg1fxx9915nqyfv1928q3k1g0-cudatoolkit-11.0.3-lib/lib:/nix/store/hcddc6qcb7mdv61qd4i5cgjz0yb43cf6-cudatoolkit-11.0-cudnn-8.1.1/lib:/nix/store/mqnx8hl70sras9szypmbn6y8bq6fcb92-nccl-2.7.8-1-cuda-11.2/lib:$(patchelf --print-rpath "/tmp/deleteme.so")" "/tmp/deleteme.so"
```
But then... it works oddly enough 😕 

But @trofi's test case looks simpler anyhow!

samuela:
Ok, I tried running this patched version of patchelf on my build, and I'm still seeing the same error. So either there's a bug in this PR, or I'm experience a separate failure mode altogether. I find the bug situation more likely since the error message is exactly the same as what's reported here.

Here's a branch with a full reproduction: https://github.com/samuela/nixpkgs/tree/samuela/pr-256-does-not-work. Just run `nix-build -A python3Packages.tensorflowWithCuda` to see the failure. Beware that building tensorflow is not for the faint of heart. I would recommend leveraging existing builds from my public cachix cache: https://app.cachix.org/cache/ploop. This way you can avoid having to build tensorflow from scratch entirely.

Mic92:
> I think the problem is not in original binary size, but in limited growth that `patchelf` allows for. We can pick a tiny hello world and try to add 32MB of data to it. Say, 32MB of NEEDED libraries:
> 
> ```
> $ nix-shell -p patchelf
> 
> # very clumsy way to achieve many long "--add-needed AAAAAAAAA...." strings.
> [nix-shell:/tmp]$ echo 'int main(){}' | gcc -x c - -o a -fPIC -pie; ./a; while patchelf --add-needed $(perl -e 'print "a" x 100000;') --add-needed $(perl -e 'print "b" x 100000;') --add-needed $(perl -e 'print "c" x 100000;') --add-needed $(perl -e 'print "d" x 100000;') --add-needed $(perl -e 'print "e" x 100000;') --add-needed $(perl -e 'print "f" x 100000') --add-needed $(perl -e 'print "g" x 100000') --add-needed $(perl -e 'print "h" x 100000') --add-needed $(perl -e 'print "i" x 100000') --add-needed $(perl -e 'print "j" x 100000') --add-needed $(perl -e 'print "k" x 100000') --add-needed $(perl -e 'print "l" x 100000') --add-needed $(perl -e 'print "m" x 100000') --add-needed $(perl -e 'print "m" x 100000') --add-needed $(perl -e 'print "o" x 100000') --add-needed $(perl -e 'print "p" x 100000') --add-needed $(perl -e 'print "q" x 100000') --add-needed $(perl -e 'print "r" x 100000') --add-needed $(perl -e 'print "s" x 100000') --add-needed $(perl -e 'print "t" x 100000') a; do echo again; done
> again
> again
> again
> again
> again
> again
> again
> again
> again
> again
> again
> again
> again
> again
> again
> again
> again
> patchelf: maximum file size exceeded
> ```

Thanks. I try to add this test case as soon as I can.

andrewla:
To be honest I'm surprised that the test works, because each resize is for the same amount, which is less that 32MB. I guess the shifting or alignment makes it a cumulative effect.

Regardless, it takes a long time for this to run, and the produced executable doesn't run because the needed libraries aren't found.

If we use add-rpath instead, we get the same effect, but we can only add one rpath per command line so it's a little less efficient.

Considering a patch that would allow text arguments to be prefixed with @ to indicate that the argument should be read from a file instead of directly from the command line; that way we can just pass in a 40MB rpath string in one go.

andrewla:
Created #316 for the `@file` syntax. That makes testing this much simpler, and is also nice because these rpaths can get crazy.

Mic92:
The `@`-pr was merged.

andrewla:
Added a test that attempts to grow the file. Have confirmed that this fails on the current master.

Mic92:
We also build now for clang for regression testing. I fixed the tests for that.

PR #261, opened 2021-01-25
Document how to find a good interpreter on NixOS

raboof:
Of course we could either have a patchelf section in the NixOS
docs or a Nix section in the patchelf docs. Since patchelf is
under the NixOS organization anyway this seemed reasonable to
me.


edolstra:
I think this is more suited to the Nixpkgs manual (especially since there is a lot more that can be said about patching 3rd party binaries).

Mic92:
I also think this better fits into NixOS documentation. For example we have https://nixos.wiki/wiki/Packaging/Binaries as a starter.
patchelf is used in a lot of other places besides Nix now.

Issue #263, opened 2021-02-17
patchelf: section header table out of bounds


asavah:
**Describe the bug**

```
patchelf --debug --set-rpath /home/asavah/kross/host/lib anyelffile
patching ELF file 'anyelffile'
patchelf: section header table out of bounds
```

**Steps To Reproduce**

see above

**Expected behavior**

rpath changed without error

patchelf built from git at https://github.com/NixOS/patchelf/commit/0ad9548c02218ac1e6d0dc1a96d0bc955ea686ee

**Additional context**

It looks like https://github.com/NixOS/patchelf/commit/cedf28411112fef1b2fc0ac848d88991efa06c87 messed up the parenthesis and changed the logic

before:

`(rdi(hdr->e_shoff) + rdi(hdr->e_shnum) * rdi(hdr->e_shentsize)) > fileContents->size()`
A+B*C > D

after:

`(rdi(hdr->e_shoff) + rdi(hdr->e_shnum)) * rdi(hdr->e_shentsize) > fileContents->size()`
(A+B)*C > D

The following patch seems to restore the logic and fix the issue here:
```
diff --git a/src/patchelf.cc b/src/patchelf.cc
index 50eb13a..9b75b6a 100644
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -397,13 +397,13 @@ ElfFile<ElfFileParamNames>::ElfFile(FileContents fileContents)
     if (rdi(hdr->e_type) != ET_EXEC && rdi(hdr->e_type) != ET_DYN)
         error("wrong ELF type");

-    if (size_t(rdi(hdr->e_phoff) + rdi(hdr->e_phnum)) * rdi(hdr->e_phentsize) > fileContents->size())
+    if (size_t(rdi(hdr->e_phoff) + rdi(hdr->e_phnum) * rdi(hdr->e_phentsize)) > fileContents->size())
         error("program header table out of bounds");

     if (rdi(hdr->e_shnum) == 0)
         error("no section headers. The input file is probably a statically linked, self-decompressing binary");

-    if (size_t(rdi(hdr->e_shoff) + rdi(hdr->e_shnum)) * rdi(hdr->e_shentsize) > fileContents->size())
+    if (size_t(rdi(hdr->e_shoff) + rdi(hdr->e_shnum) * rdi(hdr->e_shentsize)) > fileContents->size())
         error("section header table out of bounds");

     if (rdi(hdr->e_phentsize) != sizeof(Elf_Phdr))
[...]

PR #266, opened 2021-02-25
fix wrong cast

neheb:
The original code which worked was size_t(A + B * C). size_t(A + B)
breaks it. A + size_t(B * C) is correct.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

Thank you!

so that the quality of future releases can be perserved.


neheb:
Fixes: https://github.com/NixOS/patchelf/issues/263

edolstra:
Thanks!

PR #268, opened 2021-03-11
Set exact size of new load segment in executables

ehmry:
This fixes a bug where the length of new load segments did not account for the offset of the segment within the load region.

---

The problem is that the offset of the load segment created to fit a section at the beginning of the file is aligned to the start of the file, but the offset of the section in the segment is shifted by the  displacement of the ELF headers, and this is not accounted for in the size of the load segment.

The new segment could also share the same offset of the sections (preventing the ELF headers from being loading in memory), but this the most minimal fix I could find. In the file that I had a problem with, the new load segment is page-contiguous with the next segment, so the new segment could be avoided if the initial segment were expanded forwards. I'd like to implement this section merging, but not until this is fixed.

The details of the file that I had problems with:

Original
```
Section Headers:
[Nr] Name                 Type         Addr             Off      Size     ES Flags Lk Inf Al
[ 6] .dynstr              STRTAB       0000000001097bd8 00098bd8 00001e46  0 A      0   0  1
…

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  INTERP         0x096d94 0x0000000001095d94 0x0000000001095d94 0x00f634 0x00f634 R E 0x10
  LOAD           0x001000 0x0000000001000000 0x0000000001000000 0x0a8a44 0x0a8a44 R E 0x1000
  LOAD           0x0aa000 0x00000000010a9000 0x00000000010a9000 0x00f9ca 0x02ea68 RW  0x1000
…
```

The bad patch:
```
Section Headers:
[Nr] Name                 Type         Addr             Off      Size     ES Flags Lk Inf Al
[ 1] .dynstr              STRTAB       0000000000ffd190 00000190 00001fa8  0 A      0   0  8
 …

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x000000 0x0000000000ffd000 0x0000000000ffd000 0x002000 0x002000 RW  0x1000
  LOAD           0x003000 0x0000000001000000 0x0000000001000000 0x0a8a44 0x0a8a44 R E 0x1000
…
```

The good patch:
```
Section Headers:
[Nr] Name                 Type         Addr             Off      Size     ES Flags Lk Inf Al
[ 1] .dynstr              STRTAB       0000000000ffd190 00000190 00001fa8  0 A      0   0  8
…

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x000000 0x0000000000ffd000 0x0000000000ffd000 0x002138 0x002138 RW  0x1000
  LOAD           0x003000 0x0000000001000000 0x0000000001000000 0x0a8a44 0x0a8a44 R E 0x1000
…
```

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -581,18 +581,6 @@ void ElfFile<ElfFileParamNames>::shiftFile(unsigned int extraPages, Elf_Addr sta
             wri(phdrs[i].p_align, getPageSize());
         }
     }
-
-    /* Add a segment that maps the new program/section headers and
-       PT_INTERP segment into memory.  Otherwise glibc will choke. */
-    phdrs.resize(rdi(hdr->e_phnum) + 1);
-    wri(hdr->e_phnum, rdi(hdr->e_phnum) + 1);
-    Elf_Phdr & phdr = phdrs[rdi(hdr->e_phnum) - 1];
-    wri(phdr.p_type, PT_LOAD);
-    wri(phdr.p_offset, 0);
-    wri(phdr.p_vaddr, wri(phdr.p_paddr, startPage));
-    wri(phdr.p_filesz, wri(phdr.p_memsz, shift));
-    wri(phdr.p_flags, PF_R | PF_W);
-    wri(phdr.p_align, getPageSize());
 }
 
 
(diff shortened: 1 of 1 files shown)


domenkozar:
@ehmry is there a way we could have a regression test?

ehmry:
I no longer experience this problem and don't have time to properly test the fix.

PR #243, opened 2020-09-22
Corruption with program headers

eytanaim:
This corruption doesn't affect runtime linking but can be observed with inspection tools like ldd and readelf (probably because dynamic linker doesn't really go over all the program headers of shared object - only a theory)

Please use the following library for reproduction: [libm.so.6.zip](https://github.com/NixOS/patchelf/files/5259934/libm.so.6.zip)

```
eytannaim@X270-172806:/mnt/c/work/git/patchelf$ cp ../../../tmp/libm.so.6  ./
eytannaim@X270-172806:/mnt/c/work/git/patchelf$ ./src/patchelf --add-needed libpthread.so.0  ./libm.so.6
eytannaim@X270-172806:/mnt/c/work/git/patchelf$ LD_PRELOAD=./libm.so.6 sleep 1    --------> WORKS!!!!
eytannaim@X270-172806:/mnt/c/work/git/patchelf$ ldd ./libm.so.6
ldd: exited with unknown exit code (139)
eytannaim@X270-172806:/mnt/c/work/git/patchelf$ readelf  -a ./libm.so.6 > /dev/null
readelf: Error: the PHDR segment is not covered by a LOAD segment
```

The root cause of this issue is similar to the one of https://github.com/NixOS/patchelf/issues/224 or https://github.com/NixOS/patchelf/issues/239 
Thank you!

PR #278, opened 2021-05-20
Added test file for PR243, rebased off of PR270 & PR243

satmandu:
- Based upon the changes mentioned in https://github.com/NixOS/patchelf/pull/243#issuecomment-843164662

(I don't see source for the reproducer binary in https://github.com/NixOS/patchelf/pull/243#issuecomment-843164662 but perhaps the author of that comment can point that out.)


src/patchelf.cc
```diff
@@ -1057,7 +1060,7 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
 
     /* If there is a segment for the program header table, update it.
        (According to the ELF spec, there can only be one.) */
-    for (auto phdr : phdrs) {
+    for (auto &phdr : phdrs) {
```
satmandu, line 1063: Not sure if this should be `for (auto & phdr : phdrs) {` ...

satmandu:
I made the change to `for (auto & phdr : phdrs) { .` as that is used elsewhere in `patchelf.cc`. The reproducer still passes.

PR #295, opened 2021-08-09
tests/contiguous_note_sections.s: use alternate section type prefix

### LINKED ISSUE #294: v0.13 failing to build on Fedora armv7hl
Version 0.13 is failing to build in the Fedora automatic update monitoring build for the target armv7hl. It seems to build fine on the other targets. The problem is with one of the self tests (see https://kojipkgs.fedoraproject.org//work/tasks/461/73370461/build.log for more details):

```
gcc  -O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS    -march=armv7-a -mfpu=vfpv3-d16 -mtune=generic-armv7-a -mabi=aapcs-linux -mfloat-abi=hard -c -o contiguous_note_sections.o contiguous_note_sections.s
contiguous_note_sections.s: Assembler messages:
contiguous_note_sections.s:5: Error: junk at end of line, first unrecognized character is `,'
contiguous_note_sections.s:15: Error: junk at end of line, first unrecognized character is `,'
```

I'm not that familiar with using the assembler but I assume it's something to do with the `@note` following
```
.section ".note.my-section0", "a", @note
```

Looking at the GNU as manual, this seems ok to me, so it's unclear why this should be a problem. I don't have access to an armv7hl machine, so I thought I would file this in case someone knows anything before trying to make a VM.

ius:
Fixes #294

Testcase `contiguous_note_sections.sh` passes cross compiled on `armv7l-hf-multiplatform`, should ideally be verified on native ARMv6/7.

cc @jeremysanders @ovpanait

### DIFF
--- a/tests/contiguous_note_sections.s
+++ b/tests/contiguous_note_sections.s
@@ -2,7 +2,7 @@
  * Testcase for error:
  * patchelf: cannot normalize PT_NOTE segment: non-contiguous SHT_NOTE sections
  */
-.section ".note.my-section0", "a", @note
+.section ".note.my-section0", "a", %note
     .align 4
     .long 1f - 0f              /* name length (not including padding) */
     .long 3f - 2f              /* desc length (not including padding) */
@@ -12,7 +12,7 @@
 2:  .long 1                    /* desc - toolchain version number, 32-bit LE */
 3:  .align 4
 
-.section ".note.my-section1", "a", @note
+.section ".note.my-section1", "a", %note
     .align 8
     .long 1f - 0f              /* name length (not including padding) */
     .long 3f - 2f              /* desc length (not including padding) */


ovpanait:
I couldn't find any documentation indicating that "%note" syntax is valid on architectures other than arm. The GNU assembler manual (https://sourceware.org/binutils/docs-2.37/as.html) only states:

```
The optional type argument may contain one of the following constants: 

...

@note

    section contains data which is used by things other than the program 

...

Note on targets where the @ character is the start of a comment (eg ARM) then another character is used instead. For example the ARM port uses the % character. 
```

However, upon testing it seems that "%note" does compile successfully with gcc on some architectures I had access to (mips, x86_64, arm, powerpc). Not sure about toolchains other than gcc.

ius:
> I couldn't find any documentation indicating that "%note" syntax is valid on architectures other than arm.

You're right, it's undocumented. Source indicates it's fine though (see my last comment on #294).

> Not sure about toolchains other than gcc.

Fair point. I suppose this might not work for LLVM toolchains?

Mic92:
A quick test shows that llvm's assembler does not even parse the comment of the file.

```
[nix-shell:~/git/patchelf]$ llvm-as contiguous_note_sections.s
llvm-as: contiguous_note_sections.s:1:1: error: expected top-level entity
/*
^
```

I think it's fair to say that we only support binutils's gas here since llvm asm is not compatible with gas in many ways.

Mic92 (approved):
Looks good to me @domenkozar  please merge.

PR #298, opened 2021-08-10
Cleanups

Mic92:


flake.nix
```diff
@@ -27,6 +27,7 @@
         patchelf-new = final.stdenv.mkDerivation {
           name = "patchelf-${self.hydraJobs.tarball.version}";
           src = "${self.hydraJobs.tarball}/tarballs/*.tar.bz2";
+          nativeBuildInputs = [ pkgs.autoreconfHook ];
```
edolstra, line 30: Why is `autoreconfHook` needed? The tarball contains the generated configure scripts right?

Mic92: The main use case is having autotools in `nix develop` but it would also regenerate the configure script to a newer version

PR #301, opened 2021-08-18
use memcpy instead of strcpy to set rpath

Mic92:
Since we already no the size, this is faster.


Mic92:
It also gives contributor less code to copy `strcpy` to other places like here: https://github.com/NixOS/patchelf/pull/159

PR #307, opened 2021-08-25
increase test coverage

Mic92:
let's start with the easy targets from coverage reports


Mic92:
This patch also looks quite harmless as it does not add any features.

PR #309, opened 2021-08-26
add undefined behavior sanitizer

Mic92:
in addition to address sanitizer the undefined behavior sanitizer also catches bugs like integer overflows.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -58,6 +58,9 @@
           };
 
         build = forAllSystems (system: nixpkgsFor.${system}.patchelf-new);
+        build-sanitized = forAllSystems (system: nixpkgsFor.${system}.patchelf-new.overrideAttrs (old: {
+          configureFlags = [ "--with-asan " "--with-ubsan" ];
+        }));
 
         release = pkgs.releaseTools.aggregate
           { name = "patchelf-${self.hydraJobs.tarball.version}";
@@ -65,6 +68,8 @@
               [ self.hydraJobs.tarball
                 self.hydraJobs.build.x86_64-linux
                 self.hydraJobs.build.i686-linux
+                self.hydraJobs.build-sanitized.x86_64-linux
+                self.hydraJobs.build-sanitized.i686-linux
               ];
             meta.description = "Release-critical builds";
           };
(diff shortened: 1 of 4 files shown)


src/Makefile.am
```diff
@@ -1,7 +1,13 @@
 AM_CXXFLAGS = -Wall -std=c++11 -D_FILE_OFFSET_BITS=64
 
+SAN_FLAGS = -fno-optimize-sibling-calls -fsanitize-address-use-after-scope -fno-omit-frame-pointer -g -O1
```
Mic92, line 3: @andir thoughts?

src/Makefile.am
```diff
@@ -1,7 +1,20 @@
...
 if WITH_ASAN
-AM_CXXFLAGS += -fsanitize=address -fno-optimize-sibling-calls -fsanitize-address-use-after-scope -fno-omit-frame-pointer -g -O1
+AM_CXXFLAGS += -fsanitize=address
+endif
+
+if WITH_UBSAN
+AM_CXXFLAGS += -fsanitize=undefined
+endif
+
+SAN_FLAGS = -fno-optimize-sibling-calls -fsanitize-address-use-after-scope -fno-omit-frame-pointer
```
Mic92, line 11: Turns out that `-g -O2` is already part of the debug flags, so adding it here has no effect.

flake.nix
```diff
@@ -58,13 +58,18 @@
           };
 
         build = forAllSystems (system: nixpkgsFor.${system}.patchelf-new);
+        build-sanitized = forAllSystems (system: nixpkgsFor.${system}.patchelf-new.overrideAttrs (old: {
```
Mic92, line 61: @Luis-Hebendanz any other sanitizer we should add?

Luis-Hebendanz: Ah yes I would add "-fsanitize=integer" too to catch unsigned integer overflow and integer implicit unsigned integer truncation :)

Mic92: Oh. Is this not covered yet by the undefined behavior sanitizer?

Luis-Hebendanz: No because unsigned integer overflow & truncation is not undefined behavior :)

PR #311, opened 2021-08-28
make modify rpath function smaller

Mic92:
the function is still quite large but hopefully now a bit easier to follow


src/patchelf.cc
```diff
@@ -1470,11 +1485,6 @@ void ElfFile<ElfFileParamNames>::modifyRPath(RPathOp op,
     if (rpath && rpath == newRPath) {
         return;
     }
-
-    if (op == rpAdd) {
-        newRPath = std::string(rpath ? rpath : "") + ":" + newRPath;
```
Mic92, line 1475: I think this also fixes a bug where `rpAdd` would have no effect if the above condition would return early.

PR #316, opened 2021-09-10
Allow string arguments to be read from files.

andrewla:
If a string argument to an option is prefixed with @, then a file with the
following name will be read into that string.

Thank you!


andrewla:
This relates (tangentially) to #256, as testing that fix gets much simpler with this functionality. I think this change has enough merits in its own right, though.

PR #318, opened 2021-09-11
make updating .rel.plt section optional

Mic92:
I am not going to merge unless someone that has issues with updating this can confirm that the resulting binary still works.


Mic92:
Closing because lack of response for this.

PR #325, opened 2021-09-26
tests: add testcase for PT_PHDR VirtAddr corruption

ovpanait:
A PT_PHDR corruption was previously reported and fixed in [1]: the issue was
that the VirtAddr field of the PT_PHDR program header would get overwritten
with the file offset of the program header table rather than the virtual
address. A testcase for this was also added in [2]. However, the tescase is not
included in the Makefile.am regression testsuite and also tries to run a x86_64
prebuilt binary unconditionally, which would not work on other architectures.

To fix this, create a standalone testcase for the PT_PHDR VirtAddr field
corruption and include it in Makefile.am. In order to reproduce [1], a binary
with the following characteristics is needed:
- the ELF file type must be ET_DYN
- the ELF file must contain a PT_PHDR program header
- the file offset and the VirtAddr field of the PT_PHDR program header must be
  different

[1] https://github.com/NixOS/patchelf/pull/243
[2] https://github.com/NixOS/patchelf/commit/8f94e116f3c63afe224f1ec48ccad6f068f835df

Signed-off-by: Ovidiu Panait <ovpanait@gmail.com>

Thank you!


Mic92:
Thanks. This is a lot better.

PR #333, opened 2021-11-15
support loongarch for patchelf

zhangwenlong8911:
Signed-off-by: zhangwenlong <zhangwenlong@loongson.cn>

Thank you!


Mic92:
Looks like this needs an `ifdef` in the header in case `EM_LOONGARCH` is not defined.

zhangwenlong8911:
Thank you very much for your suggestion, I have modified the code according to your suggestion @Mic92

Mic92:
Thanks!

PR #337, opened 2021-11-27
Fix build-sanitized-clang.aarch64-linux build

Mic92:
We were passing stdenv for the wrong architecture


Mic92:
see https://hydra.nixos.org/build/152776211

PR #349, opened 2021-12-04
Add build & test of static musl binaries to the publish workflow

### LINKED ISSUE #330: Provide static compiled patchelf
We could statically compile patchelf with musl and upload it on github. 
This would be useful for cases that don't have patchelf in their linux distribution (see https://github.com/NixOS/patchelf/issues/329)

mayeut:
This adds build & test of static musl binaries to the publish workflow.

fixes #330

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,12 +1,15 @@
 name: Publish
 on:
+  pull_request:
   push:
+    branches:
+      - '*'
     tags:
       - '*'
 
 jobs:
-  build:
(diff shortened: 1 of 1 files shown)


Mic92:
Usually I would have build some cross-toolchain with nix but this workflow looks very simple and covers many architectures. So I would just go for this one. Also it improves cross-architecture testing.

Mic92:
Thanks!

Mic92:
There we go: https://github.com/NixOS/patchelf/releases/tag/0.14.3

PR #351, opened 2021-12-06
Use correct double-dashes for ld(1)'s build-id option

klemensn:
(Old) linkers such as BFD's ld(1) aka. GNU ld (version 2.17 on OpenBSD)
don't know `--build-id`, but they know `-b input-format` which patchelf
triggers by ommitting the second dash.

LLD's ld(1) seems to cope with this, although it's incorrect usage.

Use the double dash for correctness to get accurate error messages with
such incompatible linkers:

	-/usr/bin/ld: invalid BFD target `uild-id'
	+/usr/bin/ld: unrecognized option '--build-id'
	+/usr/bin/ld: use the --help option for usage information

Thank you!


Mic92:
Thanks!

PR #353, opened 2021-12-06
Use absoloute paths to linker scripts

klemensn:
As is common with autotools enabled software, building and testing
happens outside the source directory.

The introduction of `contiguous-note-sections` test in 0.13 broke this
by assuming linker scripts to exist in the current working directory.

Point at the absoloute source directory explicitly to unbreak linking in
such build environments.

Spotted on OpenBSD where source and build assets are commonly
separated iff possible.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -137,9 +137,9 @@ no_rpath_SOURCES = no-rpath.c
 no_rpath_CFLAGS =
 
 contiguous_note_sections_SOURCES = contiguous-note-sections.s contiguous-note-sections.ld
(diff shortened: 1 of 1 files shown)


klemensn:
With the build fixed, running tests now only shows `endianness` not being
out-of-tree aware:

```
$ cat endianness.sh.log
cp: endianness/ppc64/main: No such file or directory
cp: endianness/ppc64/libtest.so: No such file or directory
FAIL endianness.sh (exit status: 1)
```

I'll give this one a try as well when time allows.

The submitted fix lets me run tests in the first place, so that's a start.

Maybe CI infrastructure can be changed to build out-of-tree as well to catch
such behaviour?

PR #355, opened 2021-12-07
llvm/openbsd fixes

Mic92:


tests/too-many-strtab2.s
```diff
@@ -0,0 +1,5 @@
+/*
+ * Create additional .shstrtab section
+ */
+.section ".shstrtab", "", @3
```
Mic92, line 4: Mhm. This is actually not creating an additional section as intended.

Mic92: This works now.

PR #356, opened 2021-12-07
Avoid GNU seq(1)

klemensn:
MacOS, FreeBSD and Linux using GNU have it, but at least OpenBSD
does not (the non-compatible BSD equivalent jot(1) exists, though).

Counting up is easy enough to do in POSIX sh(1).

Looks like this after GNU make(1)'s escaping and produces the same
big-dynstr.c file:
```
cat main.c > big-dynstr.c
i=1; while [ $i -le 2000 ]; do echo "void f$i(void) { };"; i=$(($i + 1)); done >> big-dynstr.c
```

This is the last bit required to build and pass all tests on OpenBSD
without local patches/dependencies.

Thank you!

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -83,7 +83,7 @@ main_scoped_LDFLAGS = $(LDFLAGS_local)
 
 big-dynstr.c: main.c
 	cat $< > big-dynstr.c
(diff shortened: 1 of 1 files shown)


klemensn:
This is in light of a4027142bf7d36d983fb5286b982b038449c6750 which adequately
removes another dependency on GNU coreutils/head(1) I have already been
patching out in OpenBSD's patchelf port.

PR #358, opened 2021-12-18
print-interpreter: fix off by one error

fzakaria:
Fix off by one error in the code that reads interpreter from the ELF file. 

This was not evident when it was written directly to STDOUT but became problematic through my exploration of new functionality (#357) since there was an additional '\0' and the strings would not concatenate as a result.

```bash
> ./src/patchelf --print-interpreter /tmp/ruby
/nix/store/563528481rvhc5kxwipjmg6rqrl95mdx-glibc-2.33-56/lib/ld-linux-x86-64.so.2
```

I would love to add a test here to demonstrate the additional '\0', if there is a TODO issue for a test suite please link this issue to it.

CC @Mic92

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1237,7 +1237,7 @@ template<ElfFileParams>
 std::string ElfFile<ElfFileParamNames>::getInterpreter()
 {
     auto shdr = findSection(".interp");
-    return std::string((char *) fileContents->data() + rdi(shdr.sh_offset), rdi(shdr.sh_size));
+    return std::string((char *) fileContents->data() + rdi(shdr.sh_offset), rdi(shdr.sh_size) - 1);
 }
 
 template<ElfFileParams>
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,7 @@ Makefile
 /tests/libbig-dynstr.debug
 /tests/contiguous-note-sections
 /tests/simple-pie
+
+.direnv/
+.vscode/
+.idea/


src/patchelf.cc
```diff
@@ -1237,7 +1237,7 @@ template<ElfFileParams>
 std::string ElfFile<ElfFileParamNames>::getInterpreter()
 {
     auto shdr = findSection(".interp");
-    return std::string((char *) fileContents->data() + rdi(shdr.sh_offset), rdi(shdr.sh_size));
+    return std::string((char *) fileContents->data() + rdi(shdr.sh_offset), rdi(shdr.sh_size) - 1);
```
fzakaria, line 1240: I am trying to understand why the additional 1 here was an additional null character.

fzakaria:
As an example to demonstrate the bug I have the following change applied (and removed the fix):
```
diff --git a/src/patchelf.cc b/src/patchelf.cc
index eaf2a42..4b8436d 100644
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1237,7 +1237,7 @@ template<ElfFileParams>
 std::string ElfFile<ElfFileParamNames>::getInterpreter()
 {
     auto shdr = findSection(".interp");
-    return std::string((char *) fileContents->data() + rdi(shdr.sh_offset), rdi(shdr.sh_size) - 1);
+    return std::string((char *) fileContents->data() + rdi(shdr.sh_offset), rdi(shdr.sh_size));
 }
 
 template<ElfFileParams>
@@ -1852,8 +1852,10 @@ static bool noDefaultLib = false;
 template<class ElfFile>
 static void patchElf2(ElfFile && elfFile, const FileContents & fileContents, const std::string & fileName)
 {
-    if (printInterpreter)
-        printf("%s\n", elfFile.getInterpreter().c_str());
+    if (printInterpreter) {
+        std::string interpreter = elfFile.getInterpreter() + " you should see this";
+        printf("%s\n", interpreter.c_str());
+    }
 
     if (printSoname)
         elfFile.modifySoname(elfFile.printSoname, "");
```

Here is the output:
```
./src/patchelf --print-interpreter /tmp/ruby
/nix/store/563528481rvhc5kxwipjmg6rqrl95mdx-glibc-2.33-56/lib/ld-linux-x86-64.so.2
```
With the fix applied:
```
./src/patchelf --print-interpreter /tmp/ruby
/nix/store/563528481rvhc5kxwipjmg6rqrl95mdx-glibc-2.33-56/lib/ld-linux-x86-64.so.2 you should see this
```

src/patchelf.cc
fzakaria: Ah okay, when you construct a `std::string` from a `char *` you *should not* take into account the `\0` in the length.

Mic92:
Thanks!

PR #362, opened 2021-12-22
Add support to build with musl

fzakaria:
Add support to the flake.nix to build with musl

```
nix build .#patchelf-musl

ld-musl-x86_64.so.1 --list ./result/bin/patchelf
	/nix/store/2lp9hm9idb2y8zxfqi4nnvxsrk4vym6d-musl-1.2.2/lib/ld-musl-x86_64.so.1 (0x7fc3d791d000)
	libstdc++.so.6 => /nix/store/b29ihq5w7rh1qga809bxsdrh2mrrvd7m-gcc-10.3.0-lib/lib/libstdc++.so.6 (0x7fc3d76f7000)
	libgcc_s.so.1 => /nix/store/b29ihq5w7rh1qga809bxsdrh2mrrvd7m-gcc-10.3.0-lib/lib/libgcc_s.so.1 (0x7fc3d76dd000)
	libc.so => /nix/store/2lp9hm9idb2y8zxfqi4nnvxsrk4vym6d-musl-1.2.2/lib/ld-musl-x86_64.so.1 (0x7fc3d791d000)

./result/bin/patchelf --help
syntax: ./result/bin/patchelf
  [--set-interpreter FILENAME]
...
```

You can also launch a _develop_ shell for **glib** and **musl**.

CC @Mic92

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -93,12 +93,8 @@
 
       devShells = forAllSystems (system:
         {
-          glibc = pkgs.mkShell {
-            buildInputs = [ self.packages.${system}.patchelf.inputDerivation];
-          };
-          musl = pkgs.pkgsMusl.mkShell {
-            buildInputs = [ self.packages.${system}.patchelf-musl.inputDerivation];
-          };
+          glibc = self.packages.${system}.patchelf;
+          musl = self.packages.${system}.patchelf-musl;
         });
 
       defaultPackage = forAllSystems (system:
--- a/flake.nix
+++ b/flake.nix
@@ -21,6 +21,10 @@
 
     {
       overlay = final: prev: {
+        patchelf-new-musl = final.pkgsMusl.callPackage ./patchelf.nix {
+          inherit version;
+          src = self;
+        };
         patchelf-new = final.callPackage ./patchelf.nix {
           inherit version;
           src = self;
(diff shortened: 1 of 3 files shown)


fzakaria:
I think there is an opportunity here to converge the **hydraJobs** here too and provide the full matrix of:

```
LIBC (MUSL vs. GLIBC) x DYNAMIC (static vs dynamic) x ASAN (on/off)
```

Might be a useful PR after this.

flake.nix
```diff
@@ -85,12 +89,27 @@
...
 
+      devShell = forAllSystems (system: self.devShells.${system}.glibc);
+
+      devShells = forAllSystems (system:
+        {
+          glibc = pkgs.mkShell {
+            buildInputs = [ self.packages.${system}.patchelf.inputDerivation];
+          };
+          musl = pkgs.pkgsMusl.mkShell {
+            buildInputs = [ self.packages.${system}.patchelf-musl.inputDerivation];
```
Mic92, line 100: Currently we don't have autotools in this shell.

flake.nix
```diff
@@ -85,12 +89,27 @@
...
+          };
```
Mic92, line 101: After switching to:

```suggestion
          glibc = self.packages.${system}.patchelf;
          musl = self.packages.${system}.patchelf-musl;
```

musl just works.

PR #363, opened 2021-12-22
Introduce Finally to help track section rewrites

fzakaria:
Similarly to a concept used in NixOS/nix codebase, introduce the `Finally` class that can be given a lambda to execute on destructor.

The desired effect is to execute some code on a scope leave, in this case the method return.

This is used to make sure we only apply `sectionRewrite` once per modification and can now support early returns in the code paths.

CC @reidpr

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -18,29 +18,20 @@
 
 #include <algorithm>
 #include <limits>
-#include <map>
-#include <memory>
-#include <set>
 #include <sstream>
 #include <stdexcept>
-#include <string>
 #include <unordered_map>
-#include <vector>
-#include <optional>
-
 #include <cassert>
 #include <cerrno>
 #include <cstdarg>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
-
 #include <fcntl.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
 
-#include "elf.h"
 #include "patchelf.h"
 
 #ifndef PACKAGE_STRING
@@ -1082,6 +1073,16 @@ std::string ElfFile<ElfFileParamNames>::getInterpreter()
 template<ElfFileParams>
 void ElfFile<ElfFileParamNames>::modifySoname(sonameMode op, const std::string & newSoname)
 {
+
+    bool soNameModified = false;
+
+    Finally finally([&]() {
+        if (soNameModified) {
+            this->rewriteSections();
+            changed = true;
+        }
+    });
+
     if (rdi(hdr()->e_type) != ET_DYN) {
         debug("this is not a dynamic library\n");
         return;
@@ -1149,17 +1150,24 @@ void ElfFile<ElfFileParamNames>::modifySoname(sonameMode op, const std::string &
         setSubstr(newDynamic, 0, std::string((char *)&newDyn, sizeof(Elf_Dyn)));
     }
 
-    changed = true;
-    this->rewriteSections();
+    soNameModified = true;
 }
 
 template<ElfFileParams>
 void ElfFile<ElfFileParamNames>::setInterpreter(const std::string & newInterpreter)
 {
+    bool interpreterChanged = false;
+
+    Finally finally([&]() {
+        if (interpreterChanged) {
+            this->rewriteSections();
+            changed = true;
+        }
+    });
+
     std::string & section = replaceSection(".interp", newInterpreter.size() + 1);
     setSubstr(section, 0, newInterpreter + '\0');
-    changed = true;
-    this->rewriteSections();
+    interpreterChanged = true;
 }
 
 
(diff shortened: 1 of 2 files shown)


fzakaria:
This is desired since I saw some pull-requests may add early returns in the codepaths.
Rather than trying to track this across all pull-requests, I thought this approach similar to what Nix does might be sensible.

Open to discussion on it :)

src/patchelf.h
```diff
@@ -1,3 +1,15 @@
+#pragma once
+
+#include <string>
+#include <functional>
+#include <vector>
+#include <optional>
+#include <map>
+#include <set>
+#include <memory>
```
fzakaria, line 9: Tried to make this header file more "correct".

src/patchelf.h
```diff
@@ -157,3 +169,15 @@ class ElfFile
       return (const Elf_Ehdr *)fileContents->data();
     }
 };
+
+/* A trivial class to run a function at the end of a scope. */
+class Finally
```
fzakaria, line 174: https://github.com/NixOS/nix/blob/3228a38e1c09289b8051b42fc0da1ab0695ca9d7/src/libutil/finally.hh#L6

src/patchelf.cc
```diff
@@ -1082,6 +1073,16 @@ std::string ElfFile<ElfFileParamNames>::getInterpreter()
 template<ElfFileParams>
 void ElfFile<ElfFileParamNames>::modifySoname(sonameMode op, const std::string & newSoname)
 {
+
+    bool soNameModified = false;
+
+    Finally finally([&]() {
```
Mic92, line 1079: I would actually prefer the old code in cases where we don't need to handle early returns
to avoid hidden control flow. This one is such a case.

src/patchelf.cc
```diff
@@ -1149,17 +1150,24 @@ void ElfFile<ElfFileParamNames>::modifySoname(sonameMode op, const std::string &
...
-    this->rewriteSections();
+    soNameModified = true;
 }
 
 template<ElfFileParams>
 void ElfFile<ElfFileParamNames>::setInterpreter(const std::string & newInterpreter)
 {
+    bool interpreterChanged = false;
+
+    Finally finally([&]() {
```
Mic92, line 1161: This one as well.

src/patchelf.cc
```diff
@@ -1222,27 +1230,42 @@ std::string ElfFile<ElfFileParamNames>::shrinkRPath(char* rpath, std::vector<std
...
     Elf_Dyn * last = dyn;
     for ( ; rdi(dyn->d_tag) != DT_NULL; dyn++) {
         if (rdi(dyn->d_tag) == DT_RPATH) {
             debug("removing DT_RPATH entry\n");
-            changed = true;
+            rpathChanged = true;
         } else if (rdi(dyn->d_tag) == DT_RUNPATH) {
             debug("removing DT_RUNPATH entry\n");
-            changed = true;
+            rpathChanged = true;
```
Mic92, line 1249: This one seems fine.

src/patchelf.cc
```diff
@@ -1321,18 +1344,18 @@ void ElfFile<ElfFileParamNames>::modifyRPath(RPathOp op,
...
         dynRunPath = nullptr;
-        changed = true;
+        rpathChanged = true;
     }
 
     if (rpath && rpath == newRPath) {
         return;
     }
-    changed = true;
+    rpathChanged = true;
```
Mic92, line 1358: This one seems fine.

src/patchelf.cc
```diff
@@ -1392,6 +1414,14 @@ void ElfFile<ElfFileParamNames>::removeNeeded(const std::set<std::string> & libs
 {
     if (libs.empty()) return;
 
+    bool neededChanged = false;
+    Finally finally([&]() {
```
Mic92, line 1418: This one seems fine, but there should be no `rewriteSections` at the end.

src/patchelf.cc
```diff
@@ -1538,15 +1576,21 @@ void ElfFile<ElfFileParamNames>::replaceNeeded(const std::map<std::string, std::
...
-    this->rewriteSections();
 }
 
 template<ElfFileParams>
 void ElfFile<ElfFileParamNames>::addNeeded(const std::set<std::string> & libs)
 {
     if (libs.empty()) return;
 
+    bool neededChanged = false;
+    Finally finally([&]() {
```
Mic92, line 1587: Can be dropped here.

src/patchelf.cc
```diff
@@ -1648,15 +1698,22 @@ void ElfFile<ElfFileParamNames>::noDefaultLib()
...
-    changed = true;
+    defaultLibChanged = true;
 }
 
 template<ElfFileParams>
 void ElfFile<ElfFileParamNames>::clearSymbolVersions(const std::set<std::string> & syms)
 {
     if (syms.empty()) return;
 
+    bool symbolVersionsChanged = false;
```
Mic92, line 1709: can be dropped here.

PR #367, opened 2022-02-20
Add --add-debug option

deadw00d:
A shared library (-shared) by default does not receive DT_DEBUG tag.
This means that when a shared library has an entry point (so that it
can be run as an executable), the debugger does not connect to it
correctly and symbols are not resolved.

--add-debug option adds DT_DEBUG tag if it not yet present to an ELF
object.

Thank you!

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1653,7 +1653,7 @@ void ElfFile<ElfFileParamNames>::noDefaultLib()
 }
 
 template<ElfFileParams>
-void ElfFile<ElfFileParamNames>::addDebug()
+void ElfFile<ElfFileParamNames>::addDebugTag()
 {
     auto shdrDynamic = findSectionHeader(".dynamic");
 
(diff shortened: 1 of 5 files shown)


src/patchelf.cc
```diff
@@ -1901,6 +1938,9 @@ int mainWrapped(int argc, char * * argv)
         else if (arg == "--no-default-lib") {
             noDefaultLib = true;
         }
+        else if (arg == "--add-debug") {
```
Mic92, line 1941: Given that your described usecase is not that common, can we rename it to:

```suggestion
        else if (arg == "--add-debug-section") {
```

Than it's clear to the reader what this flag does.

deadw00d: If I understand correctly ELF, DT_DEBUG is actually a "tag" in .dynamic "section". My proposal would be to change to --add-debug-tag. Does that work for you?

Mic92: Yes.

Mic92:
One more thing: A regression test would be nice.

deadw00d:
Ok, test added

deadw00d:
Hello, is there anything else missing to merge this PR?

Mic92:
Thanks!

PR #372, opened 2022-04-06
Add O_BINARY flag when opening files to allow compilation for Windows

JagoGyselinck:
Hello!

Thank you for maintaining this tool. We've been using it at Guardsquare for a while and have recently identified a use-case where we would also need to run it on Windows hosts.

This could be achieved by switching the compiler to MinGW-w64. However, when opening the ELF file for reading/writing the code currently does not pass the `O_BINARY` flag. If the flag is not passed on Windows, the file is opened in text mode and reading/writing may prematurely end due to misinterpretation of file contents. 

This pull request resolves this issue by:
- Defining `O_BINARY` as `0` if it isn't defined, so that it has no effect on systems that do not have the `O_BINARY` flag.
- Passing `O_BINARY` when opening the ELF file for reading/writing.

Thank you for taking the time to look at this.

Best Regards,
Jago

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -47,6 +47,7 @@
 #define PACKAGE_STRING "patchelf"
 #endif
 
+// This is needed for Windows/mingw
 #ifndef O_BINARY
 #define O_BINARY 0
 #endif
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -47,6 +47,10 @@
 #define PACKAGE_STRING "patchelf"
 #endif
 
+#ifndef O_BINARY
+#define O_BINARY 0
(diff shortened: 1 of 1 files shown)


src/patchelf.cc
```diff
@@ -47,6 +47,10 @@
 #define PACKAGE_STRING "patchelf"
 #endif
 
+#ifndef O_BINARY
```
Mic92, line 50:
```suggestion
// This is needed for Windows/mingw
#ifndef O_BINARY
```

Mic92:
Thanks. It might make sense to add mingw as a target to the CI to avoid future breakage in future.

PR #376, opened 2022-05-20
modifyRPath: return early if new and old rpath are empty

ehmry:
de3901853d49e7a607fc2bcdeb18e33875bce8ce fixed the null dereference but `modifyRPath` should still return early under this condition.

ping @Mic92


Mic92:
Good catch!

PR #378, opened 2022-06-19
add --no-sort option

a-m-joseph:
This PR adds a command line option `--no-sort` which causes patchelf
to refrain from sorting the program headers and section headers.  A
comment in the preexisting source code says that this was done only
"for neatness".

The `--no-sort` option, combined with `readelf -a` and `colordiff` is
very useful for debugging patchelf problems.  Without `--no-sort` the
diffs are not usable -- everything changes because of the sorting.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -735,14 +735,16 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
     rewriteHeaders(firstPage + rdi(hdr()->e_phoff));
 }
 
+static bool noSort = false;
 
 template<ElfFileParams>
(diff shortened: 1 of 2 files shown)


src/patchelf.cc
```diff
@@ -1833,6 +1839,7 @@ void showHelp(const std::string & progName)
   [--replace-needed LIBRARY NEW_LIBRARY]\n\
   [--print-needed]\n\
   [--no-default-lib]\n\
+  [--no-sort]\t\tDo not sort program+section headers; useful for debugging patchelf.\n\
```
Mic92, line 1842: Please also mention this flag in the manpage: https://github.com/NixOS/patchelf/blob/master/patchelf.1

PR #365, opened 2022-01-22
fix incorrect polarity on dyn_offset; closes #364

a-m-joseph:
This fixes #364 and I have verified that --set-interpreter now works correctly on mips64el-linux-gnuabin32.

However --set-rpath is still broken (produces corrupted binaries); fortunately for that task `chrpath` can be used instead (it works properly).

PR #379, opened 2022-06-19
patchelf.cc: add comment explaining calculation for DT_MIPS_RLD_MAP_REL

a-m-joseph:
This commit also includes a link to the best source I could find for
what DT_MIPS_RLD_MAP_REL is supposed to be, an exact copy-paste of the
text found there, and step-by-step algebra for deriving the formula
used by the code.  Hopefully this means that this is the end of the
DT_MIPS_RLD_MAP_REL headaches.


a-m-joseph:
I thought I had already submitted this as part of #365 but it looks like I did not.

If the comment is too noisy just go ahead and close this unmerged.

PR #381, opened 2022-06-30
Add --print-os-abi and --set-os-abi options

dmsck:
This commit adds an ability to print and change the EI_OSABI field of the ELF header.


Mic92:
Changes look good. Can you also add the option to the [manpage](https://github.com/NixOS/patchelf/blob/master/patchelf.1)
and add a test in https://github.com/NixOS/patchelf/tree/master/tests ?

dmsck:
Sorry for long delay. I somehow missed the notification from GitHub.

Here you go. I've updated the documentation and added the required test case.

Mic92:
Thanks!

PR #383, opened 2022-07-16
add compiler check for c++17

Mic92:


mohammad-akhlaghi:
Thank you very much for adding the check! 

Please note that many older systems (where users don't have root permissions) don't yet have such recent compilers. It was on one such system that I found the bug above! I had to go back to patchelf 0.9 to be able to install it there! 

PatchELF is a very low-level program, and we use it in [Maneage](https://maneage.org) to build low-level components (like the compiler!). But if we can't use PatchELF on systems with an old compiler, then we won't be able to build a more recent compiler afterwards! 

Given its low-level and fundamental nature, be usable on systems with older compilers, it would be great if you can keep PatchELF compatible with at least the C++ standard of ~20 years ago (for example C++ of 2003 in this case).

Mic92:
Sorry, I am not interested in using supporting older c++ standards. You can:

- Either use the statically compiled executable
- Cross compile patchelf on a system with a newer compiler
- Use an older version of patchelf that is compatible your compiler.

PR #398, opened 2022-08-21
Validate that sections are in bounds

endofunky:
Currently, a section's sh_offset/sh_size fields are not bounds checked. This can
cause a segfault if the .interp/.dynamic sections parsed extend beyond the end
of fie.

This commit adds the missing validation to ElfFile and if found invalid issue an
error message instead.


src/patchelf.cc
```diff
@@ -266,6 +266,10 @@ ElfFile<ElfFileParamNames>::ElfFile(FileContents fContents)
         Elf_Shdr *shdr = (Elf_Shdr *) (fileContents->data() + rdi(hdr()->e_shoff)) + i;
 
         checkPointer(fileContents, shdr, sizeof(*shdr));
+
+        if (rdi(shdr->sh_offset) + rdi(shdr->sh_size) > fileContents->size())
```
Mic92, line 270: Does the elf format not allows to have sections that only exists in memory and would therefore can point to beyond the file?

PR #423, opened 2022-10-28
Set dev version suffix in master branch

haampie:
So that patchelf built off a random commit is distinguishable from a tagged commit

Edit: dunno why nix doesn't like this. But the idea is clear...


Mic92:
The trailing newline is the issue. Fixed here: https://github.com/NixOS/patchelf/pull/427

PR #428, opened 2022-11-04
fix pagesize on ia64

Mic92:


src/patchelf.cc
```diff
@@ -318,6 +318,7 @@ unsigned int ElfFile<ElfFileParamNames>::getPageSize() const
     // requirements. There is no authoritative list of these values. The
     // current list is extracted from GNU gold's source code (abi_pagesize).
     switch (rdi(hdr()->e_machine)) {
+      case EM_IA_64:
```
Mic92, line 321: Now we are at 8 broken tests down from 11.

PR #431, opened 2022-11-05
improve tests on multiple architectures

Mic92:
Thank you!


Mic92:
I will rethink this.

PR #456, opened 2023-01-27
Add options to print, clear and set executable stack state

cgzones:
Add options the modify the state of the executable flag of the GNU_STACK program header. That header indicates whether the object is requiring an executable stack.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ Makefile
 /tests/libbig-dynstr.debug
 /tests/contiguous-note-sections
 /tests/simple-pie
+/tests/simple-execstack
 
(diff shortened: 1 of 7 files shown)


Mic92:
Just for my understanding in which cases is it useful to set this flag?

cgzones:
> Just for my understanding in which cases is it useful to set this flag?

Perhaps for debugging and testing of third party tools or the kernel in case of a stack marked as executable.
My main use case currently is to drop the flag of the rustup shipped libLLVM, see https://github.com/rust-lang/rust/issues/105783.

Mic92:
bors merge

PR #459, opened 2023-02-18
Add feature to rename dynamic symbols

brenoguim:
Hello!

This PR enhances Patchelf to be able to change the name of dynamic symbols in shared libraries and executables.

This feature is similar to what is available in https://lief-project.github.io//doc/latest/tutorials/03_elf_change_symbols.html.

This can be useful when you need to change symbols, but don't have access to the source code (or object files).

The patch grew quite long, so these are the keypoints:
It adds a swich that takes a file as argument. This file is expected to have a sequence of pairs of symbol names. So if the file contains "A B C D", the symbol "A" will be renamed to "B" and symbol "C" will be renamed to "D".

First step is to iterate on the symbol table, checking their names. If it's a symbol to be renamed, append the new name to the string table, and update the symbol entry to point to the new name.
This is quite wasteful, but it should be possible to do better in the future.

Next step is to update the hash sections. Happily, they don't change in size, just content, so we just have to update it.
- `.hash` is quite straightforward and I recommend start by reviewing that one.
- `.gnu_hash` is similar to `.hash` but adds two extra complications: it has a bloom filter and it imposes that symbols in the same buckets must be adjacent in the array of symbols.

Unfortunately, changing the order of entries in the symbol array forces us to fix three other sections:
`.gnu.version` which must be in sync with the symbol table. To solve this, I just create a map of "old" position to "new" position and reorder based on that.
`.rela.dyn` and `.rela.plt` - the location tables store indices referring to the symbol table. The old2new map is also used here.

I also tried to improve a bit of the ergonomics by adding a very lightweight `span` class to represent views over arrays (https://en.cppreference.com/w/cpp/container/span will come out in c++20). It drastically the amount of pairs "pointer and size" in the code.
Also added `getSectionSpan` and `tryGetSectionSpan` to go with them.

Testing done:
Added a test that checks the symbol table, hash table (only gnu), the version table, and the relocation tables.
I also tested it on a large software with many libraries. I renamed around 600k symbols over hundreds of shared libraries, and the software runs. That's actually how I found the missing sections to update.

### DIFF
--- a/patchelf.1
+++ b/patchelf.1
@@ -124,8 +124,10 @@ Clears the executable flag of the GNU_STACK program header, or adds a new header
 Sets the executable flag of the GNU_STACK program header, or adds a new header.
 
 .IP "--rename-dynamic-symbols NAME_MAP_FILE"
-Renames dynamic symbols. The name map file should contain two space separated
-names (old name and new name) per line.
+Renames dynamic symbols. The name map file should contain lines
+with the old and the new name separated by spaces like this:
+
+old_name new_name
 
 .IP "--output FILE"
 Set the output file name.  If not specified, the input will be modified in place.
--- a/patchelf.1
+++ b/patchelf.1
@@ -129,6 +129,8 @@ with the old and the new name separated by spaces like this:
 
 old_name new_name
 
+Symbol names do not contain version specifier that are also shown in the output of the nm -D command from binutils. So instead of the name write@GLIBC_2.2.5 it is just write.
+
 .IP "--output FILE"
 Set the output file name.  If not specified, the input will be modified in place.
 
--- a/patchelf.1
+++ b/patchelf.1
@@ -123,6 +123,10 @@ Clears the executable flag of the GNU_STACK program header, or adds a new header
 .IP "--set-execstack"
 Sets the executable flag of the GNU_STACK program header, or adds a new header.
 
+.IP "--rename-dynamic-symbols NAME_MAP_FILE"
+Renames dynamic symbols. The name map file should contain two space separated
+names (old name and new name) per line.
+
 .IP "--output FILE"
 Set the output file name.  If not specified, the input will be modified in place.
 
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -2549,18 +2549,24 @@ static int mainWrapped(int argc, char * * argv)
             renameDynamicSymbols = true;
             if (++i == argc) error("missing argument");
 
-            std::ifstream infile(argv[i]);
-            if (!infile) error(fmt("Cannot open map file ", argv[i]));
+            const char* fname = argv[i];
+            std::ifstream infile(fname);
+            if (!infile) error(fmt("Cannot open map file ", fname));
 
-            std::string from, to;
-            while (true)
+            std::string line, from, to;
+            size_t lineCount = 1;
+            while (std::getline(infile, line))
             {
-                if (!(infile >> from))
+                std::istringstream iss(line);
+                if (!(iss >> from))
                     break;
-                if (!(infile >> to))
-                    error("Odd number of symbols in map file");
+                if (!(iss >> to))
+                    error(fmt(fname, ":", lineCount, ": Map file line is missing the second element"));
                 if (symbolsToRenameKeys.count(from))
-                    error(fmt("Symbol appears twice in the map file: ", from.c_str()));
+                    error(fmt(fname, ":", lineCount, ": Name ", from, " appears twice in the map file"));
+                if (from.find('@') != std::string_view::npos || to.find('@') != std::string_view::npos)
+                    error(fmt(fname, ":", lineCount, ": Name pair contains version tag: ", from, " ", to));
+                lineCount++;
                 symbolsToRename[*symbolsToRenameKeys.insert(from).first] = to;
             }
         }
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -2563,7 +2563,7 @@ static int mainWrapped(int argc, char * * argv)
                 if (!(iss >> to))
                     error(fmt(fname, ":", lineCount, ": Map file line is missing the second element"));
                 if (symbolsToRenameKeys.count(from))
-                    error(fmt(fname, ":", lineCount, ": Name ", from, " appears twice in the map file"));
+                    error(fmt(fname, ":", lineCount, ": Name '", from, "' appears twice in the map file"));
                 if (from.find('@') != std::string_view::npos || to.find('@') != std::string_view::npos)
                     error(fmt(fname, ":", lineCount, ": Name pair contains version tag: ", from, " ", to));
                 lineCount++;
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -2134,6 +2134,7 @@ void ElfFile<ElfFileParamNames>::renameDynamicSymbols(const std::unordered_map<s
         {
             wri(dynsym.st_name, strTab.size() + extraStrings.size());
             auto& newName = it->second;
+            debug("renaming dynamic symbol %s to %s\n", name.data(), it->second.c_str());
             extraStrings.insert(extraStrings.end(), newName.begin(), newName.end() + 1);
             changed = true;
         }
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -2357,7 +2357,7 @@ void showHelp(const std::string & progName)
   [--print-execstack]\t\tPrints whether the object requests an executable stack\n\
   [--clear-execstack]\n\
   [--set-execstack]\n\
-  [--rename-dynamic-symbols NAME_MAP_FILE]\tRenames dynamic symbols. The name map file should contain two symbols (old_name new_name) per line\n\
+  [--rename-dynamic-symbols NAME_MAP_FILE]\tRenames dynamic symbols. The map file should contain two symbols (old_name new_name) per line\n\
   [--output FILE]\n\
   [--debug]\n\
   [--version]\n\
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -2013,20 +2013,10 @@ void ElfFile<ElfFileParamNames>::rebuildGnuHashTable(span<char> strTab, span<Elf
     if (versyms)
         reorderSpan(versyms, old2new);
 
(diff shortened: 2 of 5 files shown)


brenoguim:
I think the CI failed because my test is trying to patch the libstdc++ of the system because it was a link. Just a guess. I'll get to it.

Thanks for runnings the CI :)

Mic92:
@brenoguim you can add https://github.com/marketplace/actions/debugging-with-tmate to your own github fork in order to debug this (and in case you cannot reproduce this locally)

brenoguim:
Thanks @Mic92 . For some reason I thought this CI required private servers, so I didn't even realize I could run it on my fork.

I'll have to rework the tests a little bit because I was renaming all symbols from lib*c++.so and making sure patchelf still runs fine.
However there are some tests linking everything statically, so that wouldn't work. I'll create a binary+library for the test.
It makes it less extensive but maybe it could be enhanced in the future (for example to include checks for version symbols).

brenoguim:
Hey @Mic92 , the CI  is clean on my fork. It should work now.

Mic92:
new merge conflicts unfortunately...

brenoguim:
No problem. I'll get to it.

Mic92:
addressed.

brenoguim:
Oh, just saw your message now. I also fixed and force pushed

src/patchelf.cc
```diff
@@ -2111,6 +2357,7 @@ void showHelp(const std::string & progName)
   [--print-execstack]\t\tPrints whether the object requests an executable stack\n\
   [--clear-execstack]\n\
   [--set-execstack]\n\
+  [--rename-dynamic-symbols NAME_MAP_FILE]\tRenames dynamic symbols. The name map file should contain two symbols (old_name new_name) per line\n\
```
Mic92, line 2360: This also needs to be in patchelf.1

brenoguim: I'm on it.

patchelf.1
```diff
@@ -123,6 +123,10 @@ Clears the executable flag of the GNU_STACK program header, or adds a new header
 .IP "--set-execstack"
 Sets the executable flag of the GNU_STACK program header, or adds a new header.
 
+.IP "--rename-dynamic-symbols NAME_MAP_FILE"
+Renames dynamic symbols. The name map file should contain two space separated
+names (old name and new name) per line.
```
Mic92, line 128:
```suggestion
Renames dynamic symbols. The name map file should contain lines
with the old and the new name separated by spaces i.e.:

old_name new_name
```

src/patchelf.cc
```diff
@@ -2242,6 +2489,25 @@ int mainWrapped(int argc, char * * argv)
         else if (arg == "--add-debug-tag") {
             addDebugTag = true;
         }
+        else if (arg == "--rename-dynamic-symbols") {
+            renameDynamicSymbols = true;
+            if (++i == argc) error("missing argument");
+
+            std::ifstream infile(argv[i]);
```
Mic92, line 2496: Should this default to stdin if `-` is passed?

Mic92: Not so sure anymore since we have multiple file arguments and it can be difficult to keep it consistent...

src/patchelf.h
```diff
@@ -1,7 +1,22 @@
 using FileContents = std::shared_ptr<std::vector<unsigned char>>;
 
-#define ElfFileParams class Elf_Ehdr, class Elf_Phdr, class Elf_Shdr, class Elf_Addr, class Elf_Off, class Elf_Dyn, class Elf_Sym, class Elf_Verneed, class Elf_Versym
-#define ElfFileParamNames Elf_Ehdr, Elf_Phdr, Elf_Shdr, Elf_Addr, Elf_Off, Elf_Dyn, Elf_Sym, Elf_Verneed, Elf_Versym
+#define ElfFileParams class Elf_Ehdr, class Elf_Phdr, class Elf_Shdr, class Elf_Addr, class Elf_Off, class Elf_Dyn, class Elf_Sym, class Elf_Verneed, class Elf_Versym, class Elf_Rel, class Elf_Rela, unsigned ElfClass
+#define ElfFileParamNames Elf_Ehdr, Elf_Phdr, Elf_Shdr, Elf_Addr, Elf_Off, Elf_Dyn, Elf_Sym, Elf_Verneed, Elf_Versym, Elf_Rel, Elf_Rela, ElfClass
+
+template<class T>
+struct span
```
Mic92, line 7: I guess something like this should be used everywhere...

src/patchelf.h
```diff
@@ -1,7 +1,22 @@
...
+{
+    span(T* d = {}, size_t l = {}) : data(d), len(l) {}
+    span(T* from, T* to) : data(from), len(to-from) {}
+    T& operator[](std::size_t i) { return data[i]; }
```
Mic92, line 11: Would it be possible to do boundary checking as well? I tried introducing it in some places but it's not yet consistent (see my `.at()` refactorings).

brenoguim: Sure. We can make the operator[] either assert or throw, whichever you prefer.

src/patchelf.h
brenoguim: Yes, I agree. It's a very simple class that makes the code very ergonomic.

src/patchelf.h
brenoguim: I added an exception.

Mic92: Yes exception is better. I prefer it to crash over random memory corruption also in production.

src/patchelf.cc
```diff
@@ -2242,6 +2493,25 @@ int mainWrapped(int argc, char * * argv)
...
+
+            std::string from, to;
+            while (true)
+            {
+                if (!(infile >> from))
+                    break;
+                if (!(infile >> to))
+                    error("Odd number of symbols in map file");
+                if (symbolsToRenameKeys.count(from))
+                    error(fmt("Symbol appears twice in the map file: ", from.c_str()));
```
Mic92, line 2511: Can we keep it c++17 for now? https://en.cppreference.com/w/cpp/utility/format/format
patchelf gets really used in many odd corners (i.e. for bootstrapping) you wouldn't expect and people were already mad at me to that they have to now install a c++17 ready compiler...

Mic92:
Sorry for these incremental reviews, but it always takes me a bit of context switch to get back into this project...

src/patchelf.cc
brenoguim: I suppose you're talking about the `fmt` call. That's a function implemented in `patchelf.cc` some lines above, it's not the C++ format.

src/patchelf.cc
```diff
@@ -1910,6 +1933,227 @@ void ElfFile<ElfFileParamNames>::addDebugTag()
...
+    std::vector<char> extraStrings;
+    extraStrings.reserve(remap.size() * 30); // Just an estimate
+    for (auto& dynsym : dynsyms)
+    {
+        std::string_view name = &strTab[rdi(dynsym.st_name)];
+        auto it = remap.find(name);
+        if (it != remap.end())
+        {
+            wri(dynsym.st_name, strTab.size() + extraStrings.size());
+            auto& newName = it->second;
```
Mic92, line 2136:
```suggestion
            auto& newName = it->second;
            debug("renaming dynamic symbol %s to %s\n", name.data(), it->second.c_str());
```

src/patchelf.cc
```diff
@@ -1910,6 +1933,227 @@ void ElfFile<ElfFileParamNames>::addDebugTag()
...
+            extraStrings.insert(extraStrings.end(), newName.begin(), newName.end() + 1);
+            changed = true;
+        }
```
Mic92, line 2139:
```suggestion
        } else {
            debug("skip dynamic symbol %sn", name.data());
        }
```

src/patchelf.cc
```diff
@@ -2291,6 +2542,25 @@ static int mainWrapped(int argc, char * * argv)
...
+            std::ifstream infile(argv[i]);
+            if (!infile) error(fmt("Cannot open map file ", argv[i]));
+
+            std::string from, to;
+            while (true)
+            {
+                if (!(infile >> from))
+                    break;
+                if (!(infile >> to))
+                    error("Odd number of symbols in map file");
```
Mic92, line 2558: I wonder if this should be a bit more strict. Currently both formats are accepted:

```
symbol
symbol_2
```

and

```
symbol symbol_2
```

patchelf.1
```diff
@@ -123,6 +123,12 @@ Clears the executable flag of the GNU_STACK program header, or adds a new header
 .IP "--set-execstack"
 Sets the executable flag of the GNU_STACK program header, or adds a new header.
 
+.IP "--rename-dynamic-symbols NAME_MAP_FILE"
+Renames dynamic symbols. The name map file should contain lines
+with the old and the new name separated by spaces like this:
+
+old_name new_name
```
Mic92, line 130:
```suggestion
old_name new_name

Symbol names do not contain version specifier that are also shown in the output of the nm -D command from binutils. So instead of the name `write@GLIBC_2.2.5` it is just `write`.
```

Mic92:
I did now a bit more manual testing an found the following rough edges above.

src/patchelf.cc
brenoguim: Yes, perhaps it was just lazyness on my part.
Requiring symbol pairs to be in one line is better because we can error on a bad map file and say the line that is wrong.

Would you agree that we should also error if an @ is found in the symbol name and mention that this does not work with symbol versions?
I saw you added it to the help, and during development I made this mistake myself (just taking the output of NM) so it would be an useful check.

I'll get to it

brenoguim: Changed it such that we read two names per line.
If the line is missing the second element, we get:
```
patchelf: <filename>:<line>: Map file line is missing the second element
```
If we have two symbols, we error out when we find the second:
```
patchelf: <filename>:<line>: Name '<name>' appears twice in the map file
```
If either of the names contain "@", we have:
```
patchelf: <filename>:<line>: Name pair contains version tag: <first name> <second name>
```

brenoguim:
Sorry for changing the code in the middle of review. I noticed I was using the syntax for template lambdas `auto x = [] <class ER> () { ... }` and that is a C++20 thing that compilers accept in C++17 mode. Still, we want to keep it C++17.
So the last patch does that. 
I also moved the helper functions to the private section.

Mic92:
bors merge

Issue #457, opened 2023-02-01
Segfault while section headers follows program headers


fish2bird:
**Describe the bug**

patchelf may bring segfault on GoLang ELF, whose section headers follows program headers and no gap between SHdr and PHdr.

**Steps To Reproduce**

```
cat > hello-world.go <<EOF
package main
import "fmt"
func main() {
    fmt.Println("hello world")
}
EOF

go build  -buildmode=pie hello-world.go
patchelf --debug --no-sort --output ./patched1 --set-interpreter /lib64/ld-linux-x86-64.so.2 ./hello-world
patchelf --debug --no-sort --output ./patched2 --set-interpreter /lib64/ld-linux-x86-64.so.2 ./patched1
./patched2

```

**Expected behavior**

- patchelf should honor space reserved for Program headers, and move if not enough while grow.
- patchelf should avoid adding new PHdr if modify in-place possible

**`patchelf --version` output**

patchelf 0.17.2

**Additional context**

for original file， Section Headers next to program headers:

```
offsetof(PHdrs) + sizeof(PHdrs[0]) * PHdrs.size() == offsetof(SHdrs)
```

Although first patch runs okay,  but PHdrs(program headers) do overlap with SHdrs(Section Headers). 

because new header added to PHdrs

```
offsetof(PHdrs) + sizeof(PHdrs[0]) * PHdrs.size() > offsetof(SHdrs)
```

details here:

```
[root tmp]#go version
go version go1.15.7 linux/amd64

[root tmp]#readelf -e ./hello-world
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x4650a0
  Start of program headers:          64 (bytes into file)
  Start of section headers:          736 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         12
  Size of section headers:           64 (bytes)
  Number of section headers:         39
  Section header string table index: 11

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
[...]

PR #460, opened 2023-02-19
Avoid overlapping program header table with section header table #457

brenoguim:
This patch does two things to fix the #457 . One could argue they should be separate fixes. They are togethere because they were both needed for the issue in #457 

Please take all changes I make with a grain of salt. I just started learning more about ELF and this codebase, so the solution could be way off.

I observed two issues:
1- Program table overlapping with section table
The tool knows that it might need to insert an extra PT_LOAD for the sections that are relocated to the end of the file. And since growing the program table could overlap with other sections, it checks and marks these overlapping sections as replaced. This makes room for the program table to grow. However, in this case, the the program table will grow into the section table which is right after. This patch detects this scenario and moves the section table to the end (before the replaced sections).

2- Multiple PT_LOAD
Everytime a string is changed (in the issue was through --set-interpreter, but in my test uses --set-soname), the string table needs to be sent to the end of the file to be updated. To do that, the tool adds a new PT_LOAD to the program table. If you repeat this process enough times, this will corrupt the library. So this patch detects that a trailing PT_LOAD already exists and extends it.
After this patch, I was able to patch the elf a thousand times and have a runnable binary.
Unfortunately, I could not understand why adding multiple PT_LOADs eventually corrupt the binary. yet.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -799,7 +799,7 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
     bool needNewSegment = true;
     auto& lastSeg = phdrs.back();
     /* Try to extend the last segment to include replaced sections */
-    if (phdrs.size() &&
+    if (!phdrs.empty() &&
         rdi(lastSeg.p_type) == PT_LOAD &&
         rdi(lastSeg.p_flags) == (PF_R | PF_W) &&
         rdi(lastSeg.p_align) == getPageSize()) {
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -755,11 +755,16 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
             replaceSection(getSectionName(shdrs.at(i)), rdi(shdrs.at(i).sh_size));
         i++;
     }
+    bool moveHeaderTableToTheEnd = rdi(hdr()->e_shoff) < pht_size;
 
     /* Compute the total space needed for the replaced sections */
     off_t neededSpace = 0;
     for (auto & s : replacedSections)
         neededSpace += roundUp(s.second.size(), sectionAlignment);
+
+    off_t headerTableSpace = roundUp(rdi(hdr()->e_shnum) * rdi(hdr()->e_shentsize), sectionAlignment);
+    if (moveHeaderTableToTheEnd)
+        neededSpace += headerTableSpace;
     debug("needed space is %d\n", neededSpace);
 
     Elf_Off startOffset = roundUp(fileContents->size(), getPageSize());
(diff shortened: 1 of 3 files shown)


src/patchelf.cc
```diff
@@ -789,24 +794,47 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
...
-    wri(phdr.p_offset, startOffset);
-    wri(phdr.p_vaddr, wri(phdr.p_paddr, startPage));
-    wri(phdr.p_filesz, wri(phdr.p_memsz, neededSpace));
-    wri(phdr.p_flags, PF_R | PF_W);
-    wri(phdr.p_align, getPageSize());
 
+    bool needNewSegment = true;
+    auto& lastSeg = phdrs.back();
+    /* Try to extend the last segment to include replaced sections */
+    if (rdi(lastSeg.p_type) == PT_LOAD &&
```
Mic92, line 802: This would probably not a valid elf file anyway but somehow I feel safer this way :)

```suggestion
    if (phdrs.size() &&
        rdi(lastSeg.p_type) == PT_LOAD &&
```

brenoguim:
> Unfortunately, I could not understand why adding multiple PT_LOADs eventually corrupt the binary. yet.

We already knew if you rewrite it twice, the PHT overlaps with SHT.
After the fix that moves the SHT to the end, you can rewrite it about 55 times.

Each of these 55 times move the .interp section to the end and add a new LOAD. The previous loads become useless, except the first that is still needed to load the SHT.

After 55 rewrites, the PHT grows so big that it displaces the next coliding section, the .text section. So both .interp and .text sections are moved. This seems to be the problem.

In the `rewriteSectionsExecutable` the code seems to be concerned about moving a PROGBITS section, but not int the `rewriteSectionsLibrary`. Not sure what is the correct thing there.

brenoguim:
Maybe a better fix is to just drop all LOAD entries that do not encompass any section. There is no point in moving the interp section further and further away in the file, leaving a blank space in the middle.

brenoguim:
> After 55 rewrites, the PHT grows so big that it displaces the next coliding section, the .text section. So both .interp and .text sections are moved. This seems to be the problem.

Perhaps it's wrong to displace the .text section because the LOAD segment we add at the end is ReadWrite, while the .text section originally loaded as ReadExec.
I suspect we would have to displace all progbits sections and track the segment attributes such that we create matching attributes. Another option is to shift their offset away from the PHT.

tests/repeated-updates.sh
```diff
@@ -0,0 +1,39 @@
...
+    ./simple || exit 1
+done
+
+load_segments_after=$(readelf -W -l libbar.so | grep -c LOAD)
+
+###############################################################################
+# To be even more strict, check that we don't add too many extra LOAD entries
+###############################################################################
+echo "Segments before: ${load_segments_before} and after: ${load_segments_after}"
+if ((${load_segments_after} > ${load_segments_before} + 2 ))
```
Mic92, line 36: For our openbsd and freebsd friends:

```suggestion
if [ ${load_segments_after} -gt ${load_segments_before} + 2 ]
```

brenoguim: We should have a rule on "make check" that runs shellcheck :D

tests/repeated-updates.sh
```diff
@@ -0,0 +1,39 @@
+#! /bin/sh -e
+
+SCRATCH=scratch/$(basename $0 .sh)
+PATCHELF=$(readlink -f "../src/patchelf")
+
+rm -rf ${SCRATCH}
```
Mic92, line 6: Can you run shellcheck on this file? https://p.krebsco.de/1fq0m92

Mic92: I will add this to our CI later.

brenoguim: Sure. I'll do that later on tonight.

tests/repeated-updates.sh
Mic92: I would but than testing shellcheck (ghc) on some more exotic platforms (mips64, sparc64...) is not possible. I think it's good enough if it's part of the ci.

brenoguim: Yes, makes more sense.
Maybe we could have a `make premerge-check` which can be used by the CI and developers

Mic92:
bors merge

Issue #449, opened 2022-12-14
After removing DT_NEEDED, clean STRTAB.


r-hoogenboom:
After removing a DT_NEEDED with --remove-needed, the removed string still appears in STRTAB, only the reference is removed in the dynamic section.

It would be nice to add an option --clean-strtab that can be used to remove all unreferenced (orphan) strings from STRTAB. I know this is not trivial, because all references to STRTAB have to be rewritten...
Maybe print the removed strings from STRTAB when using --verbose too.

amjoseph-nixpkgs:
Possibly same root cause:

- https://github.com/NixOS/patchelf/issues/453

r-hoogenboom:
Yes, it seems the issue #453 is the same root cause: only the .dynamic is changed to remove the reference to the string, but it still leaves the orphaned string in .dynstr. My suggested cleanup of .dynstr will also solve that.
I have been looking into how to clean up the .dynstr, removing all unreferenced strings, but my C++ skills are not so great. There are quite a few sections that can refer strings in .dynstr. After removing the unreferenced string(s), all offsets that refer to strings after the removed one(s) have to be updated and then the .dynstr can be shrunk. I wonder if after shrinking .dynstr, I probably need to call 'rewrite_sections', right?
Once I have it working, I can make a pull request and you can convert some C code into more C++, agree @amjoseph-nixpkgs ?

I was thinking that .dynstr could also be 'optimized' by sharing strings like 'vsprintf' with 'sprintf' and 'printf', but apparently the GNU linker already does that! This also means that for refcounting, not only the beginning of a string in .strtab should be considered, but the whole range of the string.

PR #461, opened 2023-02-19
WIP: Implementing --clean-strtab

brenoguim:
Hi,

I started to implement a simple version of `--clean-strtab` as suggested in https://github.com/NixOS/patchelf/issues/449.
No need to review it now, is work in progress.

This is a branch from the work in https://github.com/NixOS/patchelf/pull/459 so if you want to see the relevant code, just go to the last commit.

It seems to work, but needs:
 - Further validation that I got all the relevant fields in elf specification
 - Test
 - This is really `clean-dynstr` so might need work to expand to other sections


rohoog:
I see that this pull request also does what I've done in #472. However, I don't introduce new symbols due to the symbol sharing, I just leave the sharing as-is.

brenoguim:
I was planning to enhance it to do the sharing eventually, so it would be next to perfect.
I didn't check the size of the file in my test, only the presence of a given string.
I'll take a look at your PR @rohoog , maybe the approach of not undoing the sharing is the best.

Issue #446, opened 2022-11-26
0.17.0: Segmentation fault after modifying RPATH


bastimeyer:
**Describe the bug**

Just encountered an issue with patchelf 0.17.0...

I build AppImages for my Python application, and in order to do that I'm using Python's official manylinux docker images where I copy one of the pre-built python environments and modify the RPATHs of all binaries and their dependencies using patchelf. The RPATHs get set to `$ORIGIN` (and other relative paths), so that when the AppImage's squashfs gets mounted on the user's system upon execution, Python can properly be run on unknown/arbitrary mount points. So far, this has all been working flawlessly.

The patchelf 0.17.0 upgrade however has introduced segmentation faults of modified executables after modifying their RPATH. Patchelf 0.16.1 was working fine.

When looking at the recent git commit history, there was a big change in 2cb863fb756d5005f7276ad7956e01079bffce46 in regards to the ELF header file with lots of changed constants. I have 0% knowledge of any internals here and how ELF files and dynamic linking works, but that's what stood out to me.

**Steps To Reproduce**

Here's a short BASH script for reproducing the issue in two manylinux docker containers. One with patchelf 0.16.1 and the next one right after patchelf was upgraded to 0.17.0. There are other changes included between those two image versions, but those are unrelated and the issue can also be reproduced by simply building patchelf 0.17.0 on the older image.

```bash
#!/usr/bin/env bash

IMAGES=(
  # 2022-11-14 - patchelf 0.16.1
  quay.io/pypa/manylinux2014_x86_64@sha256:005826a6fa94c97bd31fccf637a0f10621304da447ca2ab3963c13991dffa013
  # 2022-11-19 - patchelf 0.17.0
  quay.io/pypa/manylinux2014_x86_64@sha256:383c6016156c94d7dbd10696c15f2444288b99a25927239b7b024e1cc6ca6a81
)

SCRIPT=$(cat <<'EOF'
PYTHON=/opt/python/cp310-cp310/bin/python

patchelf --version
$PYTHON --version

rpath=$(patchelf --print-rpath $PYTHON)
echo "RPATH: $rpath"

# Modify the python executable:
# set a different value, so the file actually gets written
patchelf --debug --set-rpath "\$ORIGIN" $PYTHON
echo "TEMP RPATH: $(patchelf --print-rpath $PYTHON)"

# and revert it again
patchelf --debug --set-rpath "$rpath" $PYTHON
[...]

PR #462, opened 2023-02-20
Add one extra page to avoid colliding with the next page being rounded down #446

brenoguim:
I wrote my analysis on https://github.com/NixOS/patchelf/issues/446 it seems that this patch fixes the issue. It adds one extra page in the "needed space". Not because the current segment needs it, but because its possible that a following segment is unaligned (and read only).
In such situation there would be a range of addresses loaded by two segments, and the read-only wins.

There is a better way to do this (check the segment for example). But I wanted to share the patch to start a discussion.

Also, it would be very important to come up with a test for this. It would be very easy to break it again while refactoring.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -479,6 +479,8 @@ static void writeFile(const std::string & fileName, const FileContents & content
 
 static uint64_t roundUp(uint64_t n, uint64_t m)
 {
+    if (n == 0)
+        return m;
     return ((n - 1) / m + 1) * m;
 }
 


brenoguim:
Added testcase. I took the binary from the issue #446 and used https://github.com/brenoguim/patchelf/tree/breno.zero_progbits to remove all the dangerous stuff from it. It now compresses much better, and we don't have the risk of running arbitrary code. The program will simply crash if you try to run it.

tests/overlapping-segments-after-rounding.sh
```diff
@@ -0,0 +1,24 @@
...
+SCRATCH="scratch/$(basename "$0" .sh)"
+READELF=${READELF:-readelf}
+
+EXEC_NAME="overlapping-segments-after-rounding"
+
+if test "$(uname -i)" = x86_64; then
+    rm -rf "${SCRATCH}"
+    mkdir -p "${SCRATCH}"
+
+    cp "${srcdir}/${EXEC_NAME}.xz" ${SCRATCH}/
```
Mic92, line 13: Can you skip the test if `xz` is not present on a system. I will make sure it will be available in CI.

brenoguim: Yes I think so. Although, I should be present, as it is passing on the CI on my side: https://github.com/brenoguim/patchelf/tree/breno.446

The file is not really big though. We can even live with having it as a whole in the repo. I think it was 200k after unxz. Will double check.

Mic92: Yes. `stdenv` we have xz but I see many distributors and other users doing `make check` during installation.

brenoguim:
I'm pretty sure I messed this up. I didn't realize the master was merged into my PR. Might need to open another one. As soon as I find my code :)

brenoguim:
Continues at https://github.com/NixOS/patchelf/pull/469 sorry for the mess

PR #464, opened 2023-02-21
Modernizations and strictness improvements

cgzones:
* modernize the codebase by using modern C++ features
* be more const-correct by using adjusting casts and enable `-Wcast-qual`
* declare more read-only functions `const`
* avoid value truncations in multiple places
* avoid out-of-bounds memory access and integer overflows on malicious ELF data

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -243,6 +243,25 @@ static void checkPointer(const FileContents & contents, const void * p, size_t s
 }
 
 
+static void checkOffset(const FileContents & contents, size_t offset, size_t size)
+{
+    size_t end;
+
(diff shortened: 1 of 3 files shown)


src/patchelf.cc
```diff
@@ -196,11 +196,11 @@ static FileContents readFile(const std::string & fileName,
     while ((portion = read(fd, contents->data() + bytesRead, size - bytesRead)) > 0)
         bytesRead += portion;
 
+    close(fd);
```
Mic92, line 199: This could be even autoclosed later on.

Mic92:
bors merge

PR #466, opened 2023-02-22
Adjust roundUp for 0 as input

cgzones:
Round up 0 to m instead of wrapping around and return an unexpected result, which is not a multiple of m.


Mic92:
bors merge

PR #467, opened 2023-02-22
Compute program header flags in rewriteSectionsLibrary

cgzones:
Instead of using fixed flags of RW for the new program header, holding
the replaced sections, check if the write or execute flags is required
by any accompanying section.

If many sections are going to be replaced a more strict method would be
to split them up and add up-to 4 new program headers if necessary:
read-only, read-write, read-execute, read-write-execute.

p.s.:
also the harded flags for executables seems suspicious:
https://github.com/NixOS/patchelf/blob/e37f892b12ff900ab5e733f688548c87cd14ef02/src/patchelf.cc#L562
Maybe that should be `wri(phdr.p_flags, rdi(phdrs.at(splitIndex).p_flags));`?

p.p.s:
Should these lines convert the read data
https://github.com/NixOS/patchelf/blob/e37f892b12ff900ab5e733f688548c87cd14ef02/src/patchelf.cc#L558-L560
e.g.
```c

 wri(phdr.p_offset, rdi(phdrs.at(splitIndex).p_offset) - splitShift - shift); 
 wri(phdr.p_paddr, rdi(phdrs.at(splitIndex).p_paddr) - splitShift - shift); 
 wri(phdr.p_vaddr, rdi(phdrs.at(splitIndex).p_vaddr) - splitShift - shift);
```
?

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -842,7 +842,17 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
     wri(phdr.p_offset, startOffset);
     wri(phdr.p_vaddr, wri(phdr.p_paddr, startPage));
     wri(phdr.p_filesz, wri(phdr.p_memsz, neededSpace));
-    wri(phdr.p_flags, PF_R | PF_W);
+    {
+        bool need_write = false, need_exec = false;
+        for (const auto & rs : replacedSections) {
+            const auto flags = findSectionHeader(rs.first).sh_flags;
+            if (flags & SHF_WRITE)
+                need_write = true;
+            if (flags & SHF_EXECINSTR)
+                need_exec = true;
+        }
+        wri(phdr.p_flags, PF_R | (need_write ? PF_W : 0) | (need_exec ? PF_X : 0));
+    }
     wri(phdr.p_align, getPageSize());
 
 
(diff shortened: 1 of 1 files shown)


Mic92:
cc @brenoguim

src/patchelf.cc
```diff
@@ -844,7 +842,17 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
...
+    {
+        bool need_write = false, need_exec = false;
+        for (const auto & rs : replacedSections) {
+            const auto flags = findSectionHeader(rs.first).sh_flags;
+            if (flags & SHF_WRITE)
+                need_write = true;
+            if (flags & SHF_EXECINSTR)
+                need_exec = true;
+        }
+        wri(phdr.p_flags, PF_R | (need_write ? PF_W : 0) | (need_exec ? PF_X : 0));
```
brenoguim, line 854: Another way to write this would be:
```
auto flagUnion = PF_R;
for (const auto& rs : replacedSections) {
    const auto flags = findSectionHeader(rs.first).sh_flags;
    flagUnion |= (flags & SHF_WRITE) ? PF_W : 0;
    flagUnion |= (flags & SHF_EXECINSTR) ? PF_X : 0;
}
wri(phdr.p_flags, flagUnion);
```
But that's style, feel free to drop.

brenoguim (approved):
I'm glad that this is being done, because it's a bit concerning that some sections are exposed with different access rights. Even though it doesn't fix everything, it's already a good improvement.

This is the kind of bug that would fit in perfectly with https://github.com/NixOS/patchelf/issues/468 because it would be lovely have a test for it. And in the current state of things, it would be cumbersome to create a reliable one.

brenoguim:
> also the harded flags for executables seems suspicious:

Since these flags are not being tracked, RW is the "works for all". I don't think we ever move executable sessions (because they are usually progbits) so we never needed.

> Should these lines convert the read data

I think so. I would expect things to break drastically if endianess is not respected. Do you know if we have tests for that?

PR #469, opened 2023-02-24
Add one extra page to avoid overlapping with next page if its rounded…

brenoguim:
… down


tests/overlapping-segments-after-rounding.sh
```diff
@@ -0,0 +1,23 @@
+#! /bin/sh -e
+
+PATCHELF=$(readlink -f "../src/patchelf")
+SCRATCH="scratch/$(basename "$0" .sh)"
+READELF=${READELF:-readelf}
+
+EXEC_NAME="overlapping-segments-after-rounding"
+
+if test "$(uname -i)" = x86_64; then
```
Mic92, line 9: We also need to check for linux here, no?

brenoguim: I guess so, but I'm not sure I understand how this is used in windows.
If the CI tests on windows, then it would be failing.
If it doesn't, then there is no need to check.
Unless we have people running make check on windows but the CI cannot test it for some reason.

Mic92: On windows it's mostly used in cross builds i.e. for android. But I am not sure how the tests works on other elf-based operating systems. I can test in a VM later.

PR #471, opened 2023-03-02
Add define for MIPS_XHASH

LinuxUserGD:
Fixes undeclared identifier error on musl libc (Gentoo amd64/clang), see https://mail-index.netbsd.org/source-changes-hg/2021/05/01/msg290415.html
```bash
clang++ -std=c++17 -DPACKAGE_NAME=\"patchelf\" -DPACKAGE_TARNAME=\"patchelf\" -DPACKAGE_VERSION=\"0.17.2\" -DPACKAGE_STRING=\"patchelf\ 0.17.2\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DPACKAGE=\"patchelf\" -DVERSION=\"0.17.2\" -DHAVE_CXX17=1 -I.    -Wall -std=c++17 -D_FILE_OFFSET_BITS=64     -O3 -pipe -march=skylake -mtune=skylake -D_FORTIFY_SOURCE=3 -stdlib=libc++ -c -o patchelf.o patchelf.cc
patchelf.cc:1087:33: error: use of undeclared identifier 'DT_MIPS_XHASH'
            } else if (d_tag == DT_MIPS_XHASH) {
                                ^
1 error generated.
make[1]: *** [Makefile:380: patchelf.o] Error 1
make[1]: Leaving directory '/var/tmp/portage/dev-util/patchelf-0.17.2/work/patchelf-0.17.2/src'
make: *** [Makefile:448: all-recursive] Error 1
```

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -68,6 +68,10 @@ static int forcedPageSize = DEFAULT_PAGESIZE;
 static int forcedPageSize = -1;
 #endif
 
+#ifndef DT_MIPS_XHASH
+#define DT_MIPS_XHASH   0x70000036
+#endif
+
 #ifndef EM_LOONGARCH
 #define EM_LOONGARCH    258
 #endif


tgbugs:
Is it safe to define that constant if it has not already been defined? Is it a glibc specific constant or is it an implementation of something that is standard on MIPS?

LinuxUserGD:
> Is it safe to define that constant if it has not already been defined? Is it a glibc specific constant or is it an implementation of something that is standard on MIPS?

It's a constant in [binutils](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=include/elf/mips.h;h=e2c3868348a59400cd31f5d5597d2ef656540d9b;hb=HEAD#l767) so probably not specific to glibc-only systems I think

brenoguim:
> Edit: Also seems to be defined in
> 
> https://github.com/NixOS/patchelf/blob/ea2fca765c440fff1ff74e1463444dea7b819db2/src/elf.h#L2011
 
Since this is already defined, why would this PR be necessary? Perhaps the real problem is another one?

LinuxUserGD:
> > Edit: Also seems to be defined in
> > https://github.com/NixOS/patchelf/blob/ea2fca765c440fff1ff74e1463444dea7b819db2/src/elf.h#L2011
> 
> Since this is already defined, why would this PR be necessary? Perhaps the real problem is another one?

Maybe elf.h from musl libc is included first https://git.musl-libc.org/cgit/musl/tree/include/elf.h which doesn't define `DT_MIPS_XHASH` so `_ELF_H` is already defined without constants from https://github.com/NixOS/patchelf/blob/master/src/elf.h

There's also a gentoo bug https://bugs.gentoo.org/860888 removing `DT_MIPS_XHASH` which might be a better approach.

brenoguim:
I'm confused. So even though patchelf contains the elf.h file used to build and test, it needs to also work with some other elf.h? It seems backwards to me.
If the problem is indeed musl providing another elf.h (needs verification) it would be nicer if it just didn't unless we explicitly ask for it.
Let me try to reproduce the issue and better understand what is going on.

LinuxUserGD:
> I'm confused. So even though patchelf contains the elf.h file used to build and test, it needs to also work with some other elf.h? It seems backwards to me. If the problem is indeed musl providing another elf.h (needs verification) it would be nicer if it just didn't unless we explicitly ask for it. Let me try to reproduce the issue and better understand what is going on.

Ok, Gentoo actually removes removes the file which causes the build failure: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/patchelf/patchelf-0.17.2.ebuild#n17
Sry, seems like code should be fine upstream, I guess a downstream gentoo patch might be better than.

Issue #453, opened 2023-01-12
patchelf --remove-rpath leaves the rpath string the file


amjoseph-nixpkgs:
**Describe the bug**

After running `patchelf --remove-rpath` on a shared library, `readelf -a | grep $OLD_RPATH` will show no matches.  However the string has been left behind in the binary.

**Steps To Reproduce**

Please include exact steps with an attached binary so that
another person can reproduce the problem.

```
patchelf --remove-rpath libgcc_s.so.1
readelf -a libgcc_s.so.1 | grep /nix/store/kvjzg
strings libgcc_s.so.1 | grep /nix/store/kvjzg
/nix/store/7rb10j6ifizp05pm2h51sjwiky5wnkb8-xgcc-11.3.0-lib/lib:/nix/store/kvjzg5xf51niicmw9fndxws1akgz21p4-bootstrap-stage0-glibc-bootstrap/lib:/nix/store/4b30jk1h42q5qs71l4sbv8cr3hnxdys4-bootstrap-tools/lib
```

**Expected behavior**

No output from the second command.  Nix's reference scanner depends on this.

**`patchelf --version` output**

```
patchelf 0.15.0
```
[libgcc_s.so.1.gz](https://github.com/NixOS/patchelf/files/10399196/libgcc_s.so.1.gz)

**Additional context**

amjoseph-nixpkgs:
Temporary workaround is `patchelf --set-rpath ""`

Mic92:
Just now it only removes the dyn section: https://github.com/NixOS/patchelf/blob/5908e16cd562bcb1909be4de0409c4912a8afc52/src/patchelf.cc#L1400 it could also override the old string with `XXX`

jvolkman:
> Just now it only removes the dyn section:
> 
> https://github.com/NixOS/patchelf/blob/5908e16cd562bcb1909be4de0409c4912a8afc52/src/patchelf.cc#L1400
> 
> it could also override the old string with `XXX`

Isn't modifying the old value potentially dangerous? Is there anything that says multiple dyn entries can't refer to the same string in the table (or even the suffix of an existing string)?

amjoseph-nixpkgs:
> Isn't modifying the old value potentially dangerous? Is there anything that says multiple dyn entries can't refer to the same string in the table (or even the suffix of an existing string)?

Yikes.  That is probably the case.

Mic92:
> 
[...]

PR #472, opened 2023-03-02
Cleanstrtab

rohoog:
I have created the change that adds cleaning of the dynstr. All unreferenced symbols are removed if --clean-strtab option is given and the section is shrunk. It could solve issues #453 and #449.

However I found that the binary actually grows by about 4KB. I think this is due to this note section normalizing. I'm not sure why this is done and why this should be needed. Why not leave all sections before the one(s) that are modified alone?


src/patchelf.cc
```diff
@@ -1885,9 +1906,186 @@ void ElfFile<ElfFileParamNames>::clearSymbolVersions(const std::set<std::string>
...
+        return (a < b.str);
+    }
+
+    bool operator()(const symstr& a, const char* b) const
+    {
+        return (a.str+a.len < b);
+    }
+};
+
+#include <argz.h>
```
brenoguim, line 2258: This might be glibc specific. Should we rely on it?

src/patchelf.cc
```diff
@@ -1885,9 +1906,186 @@ void ElfFile<ElfFileParamNames>::clearSymbolVersions(const std::set<std::string>
...
+
+class mymap : public map<symstr, forward_list<symref>, Comparator> {
```
brenoguim, line 2260: Perhaps one would want a better name for this class

src/patchelf.cc
```diff
@@ -1885,9 +1906,186 @@ void ElfFile<ElfFileParamNames>::clearSymbolVersions(const std::set<std::string>
...
+void ElfFile<ElfFileParamNames()>::cleanstrtab()
+{
+    auto shdrDynamic = findSectionHeader(".dynamic");
+    auto shdrDynStr = findSectionHeader(".dynstr");
+    auto shdrDynSym = findSectionHeader(".dynsym");
+
+    auto strTab = (char *)fileContents->data() + rdi(shdrDynStr.sh_offset);
+    int verneednum = 0;
+    int verdefnum = 0;
+    //mkref(strTab, rdi(shdrDynStr.sh_size)); // new strInfo()
```
brenoguim: cleanup old code

src/patchelf.cc
```diff
@@ -1855,12 +1876,12 @@ void ElfFile<ElfFileParamNames>::addDebugTag()
...
     setSubstr(newDynamic, 0, std::string((char *) &newDyn, sizeof(Elf_Dyn)));
 
-    this->rewriteSections();
+    rewriteSections();
     changed = true;
 }
 
 template<ElfFileParams>
-void ElfFile<ElfFileParamNames>::clearSymbolVersions(const std::set<std::string> & syms)
+void ElfFile<ElfFileParamNames()>::clearSymbolVersions(const std::set<std::string> & syms)
```
brenoguim, line 2201: It would be nicer if these changes weren't made in this PR, or at least in a different commit

src/patchelf.cc
```diff
@@ -1885,9 +1906,186 @@ void ElfFile<ElfFileParamNames>::clearSymbolVersions(const std::set<std::string>
         }
     }
     changed = true;
-    this->rewriteSections();
+    rewriteSections();
 }
 
+struct symstr {
```
brenoguim, line 2229: Perhaps one could use a std::string_view for this?

src/patchelf.cc
```diff
@@ -1885,9 +1906,186 @@ void ElfFile<ElfFileParamNames>::clearSymbolVersions(const std::set<std::string>
...
+					auxnum--;
+				}
+				versym = (Elf_Verneed *)((char*)versym + rdi(versym->vn_next));
+				verneednum--;
+			}
+		}
+    }
+
+    if (verdefnum) {
+        auto shdrVerSymd = findSectionHeader(".gnu.version_d");
```
brenoguim, line 2346: Mixing tab and spaces. You might want to look into the identation to match the rest of the file

src/patchelf.cc
```diff
@@ -39,6 +41,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <stdlib.h>
```
brenoguim, line 47: In c++ this should be `#include <cstdlib>`

src/patchelf.cc
```diff
@@ -429,14 +434,8 @@ static void writeFile(const std::string & fileName, const FileContents & content
 }
 
 
-static uint64_t roundUp(uint64_t n, uint64_t m)
-{
-    return ((n - 1) / m + 1) * m;
-}
-
```
brenoguim, line 486: is this change related?

src/patchelf.cc
```diff
@@ -600,7 +614,7 @@ void ElfFile<ElfFileParamNames>::writeReplacedSections(Elf_Off & curOff,
         const std::string & sectionName = i.first;
         Elf_Shdr & shdr = findSectionHeader(sectionName);
         if (rdi(shdr.sh_type) != SHT_NOBITS)
-            memset(fileContents->data() + rdi(shdr.sh_offset), 'X', rdi(shdr.sh_size));
+            memset(fileContents->data() + rdi(shdr.sh_offset), 'Y', rdi(shdr.sh_size));
```
brenoguim, line 686: is this necessary?

src/patchelf.cc
```diff
@@ -712,16 +726,17 @@ void ElfFile<ElfFileParamNames>::writeReplacedSections(Elf_Off & curOff,
                 }
             }
         }
-
-        curOff += roundUp(i->second.size(), sectionAlignment);
+        debug("curoff %d, size %d, align %d\n", curOff, i->second.size(), sectionAlignment);
+        curOff += roundUp((Elf_Off)i->second.size(), sectionAlignment);
+        debug("-> next curoff %d\n", curOff);
```
brenoguim, line 800: What's the purpose of this change?

src/patchelf.cc
```diff
@@ -1514,7 +1529,7 @@ void ElfFile<ElfFileParamNames>::modifyRPath(RPathOp op,
     size_t rpathSize = 0;
     if (rpath) {
         rpathSize = strlen(rpath);
-        memset(rpath, 'X', rpathSize);
+        memset(rpath, 0, rpathSize);
```
brenoguim, line 1635: Is this change necessary?

src/patchelf.cc
```diff
@@ -1587,13 +1603,18 @@ void ElfFile<ElfFileParamNames>::removeNeeded(const std::set<std::string> & libs
             *last++ = *dyn;
     }
 
-    memset(last, 0, sizeof(Elf_Dyn) * (dyn - last));
+    if (!changed) return;
+	auto newsize = rdi(shdrDynamic.sh_size) - sizeof(Elf_Dyn) * (dyn - last);
+	wri(shdrDynamic.sh_size, newsize);
+	dynamic.resize(newsize);
+	debug("new .dynamic size %d\n", rdi(shdrDynamic.sh_size));
+	replaceSection(".dynamic", dynamic);
```
brenoguim, line 1715: Misleading indentation

brenoguim (commented):
Overall, it seems there are many changes in this PR, some needed for the functionality and some cosmetic. This makes it a bit hard to review with confidence.

Also, adding a test is pretty important, so it would go a long way.

src/patchelf.cc
rohoog: Why not?

src/patchelf.cc
rohoog: What would you suggest? symmap or such?

src/patchelf.cc
rohoog: I'll (try to) split it in cosmetic commit and functional commit.

src/patchelf.cc
rohoog: maybe... This looks most like the original C-implementation I did. However, I need none of the 'burden' of string_view.

src/patchelf.cc
rohoog: I was hoping eclipse would do that automatically, but alas it doesn't... What is the style? tabs or spaces?

src/patchelf.cc
rohoog: And the above cunistd? Sorry I come from a C background... Probably stdlib isn't needed anymore now I don't malloc() anymore.

src/patchelf.cc
rohoog: no, part of the 'cosmetic' changes.

src/patchelf.cc
rohoog: left-over from my debugging

src/patchelf.cc
rohoog: debugging left-over

src/patchelf.cc
rohoog: debugging left-over

src/patchelf.cc
rohoog: yeah, this used to be if (changed) {...

brenoguim:
I also forgot to say. You might want to enable the CI in your github fork so that you can make sure all tests are passing.

src/patchelf.cc
brenoguim: There is no cunistd, since that one is unix specific.

src/patchelf.cc
brenoguim: Spaces

brenoguim:
We are going to need to decide between my implementation and yours.
Mine has the disadvantage of undoing the sharing, but I intend to implement sharing detection on it, which would then create sharing for the modifications.
Yours persist the sharing, but would not add sharing. I suspect adding sharing detection on my implementation would be more fiting the algorithm, while in yours it would need a more significant rewrite.
What do you think about it?

src/patchelf.cc
rohoog: weird, this line is gone already here.... commit "remove some leftover comment-out lines " should have removed it.

src/patchelf.cc
brenoguim: I suspect there are tests in the CI that compile patchelf against musl libc. But maybe I'm wrong. We'll know when the CI runs.

src/patchelf.cc
rohoog: looked into string_view and the constructor string_view(char *s) can be useful to replace the argz_* functions too. Good tip,

rohoog:
Sorry, I didn't look into your implementation. Mine was a C++ learning experience for me. Do with it as you please.
I addressed your review comments, except the split of the initial big commit and the adding of the test case (if you did a test-case, then that would probably work for my code too). Please have a look.
Initially I also had the idea of adding the symbol sharing, until I found that the gnu binutils already does that, so I didn't bother. Since the 'patching' is about removing DT_NEEDED, I don't think there will be new opportunities for sym sharing.

I do find the growing of the binary disturbing. Also the number of program headers increases by the 'rewriteSections' method. Should I make a new issue for that?

brenoguim:
> Since the 'patching' is about removing DT_NEEDED, I don't think there will be new opportunities for sym sharing.

Patchelf has many operations that will grow the dynstr table. For example, changing RPATH, or renaming symbols. So redoing the sharing could improve in those cases.

> I do find the growing of the binary disturbing. Also the number of program headers increases by the 'rewriteSections' method. Should I make a new issue for that?

There is definitely room for improvement indeed. Patchelf will run a generic algorithm to change the place of modified sections, and to do that it needs to add more loads.
But if you know that you did not grow any section, you can just tune the algorithm to avoid calling rewriteSections.

brenoguim:
I understand that. I just meant that it would be useful to redo the sharing because some options in patchelf add new strings.

If you just set changed=true without rewriting the sections, the binary will have exactly the same size. At least it won't grow. Patchelf will just update the changed bytes in the file.

If you want to actually shrink, I guess this needs to be implemented.

It's complicated because if you want to keep the section in the same place as it was before and make the file smaller, that would mean all other sections would be loaded in an earlier address space, which breaks the binary.
If you want to move the new smaller section to another place in the virtual space, you need to add a new program header table entry. And maybe there isn't room for that, forcing you to displace the next section into another location. But to displace sections, you might have to normalize note sections... So you get to the issue existing today.

r-hoogenboom:
HMM, I expected that exactly this was the responsibility of rewriteSections. If it can deal with grown sections, why wouldn't it be able to deal with shrunk sections.

Note sections are just arbitrary 'advisory' non-critical things (like which assembler assembled the binary, which linker linked it, etc), right? Do they really refer to absolute locations of data in the binary?

I think patchelf is growing to be a full-blown linker (and un-linker) and by that, one that supports many architectures and bit-nesses. I think patchelf should leave stuff alone which it can, try to leave as much as possible to the linker that originally built the binary. But that's just my opinion.

src/patchelf.cc
Mic92: Yes. Musl support is needed in general we not just support Linux but also other operating systems.

r-hoogenboom: its already removed with the move to string_view.

Issue #416, opened 2022-10-13
Add option to replace an imported symbol


raphaelr:
**Is your feature request related to a problem? Please describe.**
Using the rpath arguments (e.g. `--add-rpath`), patchelf can override the location of libraries. However, libraries have other ways of depending on file system paths. A common example is drivers containing a hard-coded path to firmware binaries, which the library `fopen()`s.

Example: [Broadcom fingerprint driver](https://git.launchpad.net/~oem-solutions-engineers/libfprint-2-tod1-broadcom/+git/libfprint-2-tod1-broadcom/tree/), the .so contains a hardcoded path to `/var/lib/fprint/fw`.

Currently it is difficult to replace such paths.

**Describe the solution you'd like**
Add a new argument to patchelf, `--replace-symbol NAME_ORIG NAME_NEW`. If specified, patchelf will iterate over the symbol table (`.dynsym` section), and replace the symbol with the name *NAME_ORIG* with *NAME_NEW*, adding it to the `.dynstr` section if neccessary.

This allows one to replace the `fopen` call from the library with a wrapper function, which can inspect the path it receives and change it if neccessary.

Example invocation:
```
$ # assume that fopen_wrapper.c defines a `fopen_wrapper()` function with the same signature as `fopen()`
$ cc -fPIC -shared fopen_wrapper.c -o libfopen_wrapper.so
$ patchelf \
    --replace-symbol fopen fopen_wrapper \
    --add-needed libfopen_wrapper.so \
    libdriver.so
```

**Describe alternatives you've considered**
* Use `buildFHSUserEnv` from nixpkgs to give the process the filesystem layout it needs.
* Use LD_PRELOAD to inject the wrapper into the process.

These methods have the disadvantage that they affect an entire process tree, not just the single library that needs patching.

**Additional context**
I have a [prototype quality](https://git.sr.ht/~raphi/patchelf/commit/9079354d2b94050f38594b5fab5f764f73e4b7e0) implementation of this feature. If this feature request is accepted, I'll clean it up and add tests/documentation for it.

Upstreaming of this feature has been requested by the [iscan-snap](https://github.com/brlin-tw/iscan-snap/blob/08076b1/snap/local/launchers/iscan-launch#L30-L50) project, which attempts to package scanner drivers using snap.

Mic92:
Sorry for the late response. I just now catched up with the latest issues and PRs in the project again.
[...]

PR #473, opened 2023-03-04
Other switches might set changed as true. Use extraStrings size. #416

brenoguim:
I didn't realize that "changed" could be set by other switches before reaching the renaming function.
In this case, it tried to add zero strings and failed an assertion due to that.
Added a test that breaks without the fix.


Mic92:
bors merge

Issue #474, opened 2023-03-05
SEGFAULT with auditwheel + patchelf


magicse:
**Describe the bug**
SEGFAULT with auditwheel + patchelf 
General problem that wheel repaired in Qemu Docker image didn't work in the same image Docker on real ARMv7, and without wheelaudit+patchelf,  built packages work well.
GDB show that problem in patched .libs/libstdc++-ad7c573d.so.6.0.29 .libs/libgcc_s-8c7760c8.so.1
 
Patched in Qemu docker ARMv7 libraries after that give SEGFAULT on real ARMv7
https://github.com/pypa/auditwheel/issues/414#issue-1610229237

**Steps To Reproduce**
All steps in this post
https://github.com/home-assistant/core/issues/86589#issuecomment-1455094527

**Expected behavior**
Worked python wheel
**`patchelf --version` output**
patchelf 0.17.2
**Additional context**

brenoguim:
Would it be possible to attach de libstdc++ before and after patching? And also the patchelf command that was used? That would help a lot in the investigation.

magicse:
> Would it be possible to attach de libstdc++ before and after patching?

No problem. I'll make this

magicse:
Here is readelf logs 
[readelf-aW.zip](https://github.com/NixOS/patchelf/files/10892258/readelf-aW.zip)


> And also the patchelf command that was used?

I dont know, bacuse this commands make auditwheel python package

brenoguim:
Looks like the alignment of the existing LOAD segments is `0x10000` while patchelf is adding two new load segments with alignment `0x1000`. It seems that armv7 requires the former alignment. I don't know for sure if that is the cause, but it's a good guess.
There is code to detect the correct alignment in patchelf, let me see why it didn't kick int.

magicse:
This is already a good step towards solving the problem.

brenoguim:
@magicse would you be able to test the build from https://github.com/NixOS/patchelf/pull/475 ? I don't have an ARMv7 box.

magicse:
Yes sure

magicse:
```console
bash-5.1# ls
cchardet-2.1.7-cp310-cp310-musllinux_1_2_armv7l.whl
bash-5.1# python -c "import cchardet"
Segmentation fault
bash-5.1# pip install --force-reinstall cchardet-2.1.7-cp310-cp310-musllinux_1_2_armv7l.whl 
[...]

PR #475, opened 2023-03-05
Use the largest segment alignment for libraries requiring non-standard alignments #474

brenoguim:
It turns out libraries can have a larger than usual alignment on their LOAD segments. Patchelf should try to play along and use the largest alignment if its going to add a new LOAD segment

### DIFF
--- a/tests/set-rpath-library.sh
+++ b/tests/set-rpath-library.sh
@@ -61,7 +61,7 @@ fi
 # ALL loads should have the same alignment
 lib="${SCRATCH}/liboveralign.so"
 ../src/patchelf --set-rpath "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" "$lib"
-num_alignments=$(${READELF} -W -l "${lib}"  | grep LOAD | awk '{ print $NF }' | sort -u | wc -l)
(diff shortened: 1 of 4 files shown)


brenoguim:
The binaries in the arm32v7 CI do not have the 0x10000 alignment. They have 0x1000. So I guess there must be another way to detect what is the correct alignment to be used.
Perhaps the safest way is to use the alignment of the LOAD segment from which we are removing the section.

tests/set-rpath-library.sh
```diff
@@ -56,3 +57,12 @@ if test "$exitCode" != 46; then
     echo "bad exit code!"
     exit 1
 fi
+
+# ALL loads should have the same alignment
+lib="${SCRATCH}/liboveralign.so"
+../src/patchelf --set-rpath "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" "$lib"
+num_alignments=$(${READELF} -W -l "${lib}"  | grep LOAD | awk '{ print $NF }' | sort -u | wc -l)
```
Mic92, line 64: nitpick:

```suggestion
num_alignments=$(${READELF} -W -l "${lib}"  | awk '/LOAD/ { print $NF }' | sort -u | wc -l)
```

Mic92: otherwise looks good.

PR #481, opened 2023-03-17
Do not let modifyRPath taint shared strings in strtab. Fix #315

### LINKED ISSUE #315: patchelf --set-rpath adds external unresolved symbol XXX to binary
I'm building strongswan on Openwrt 21.02 for newport platform.
I noticed that the swanctl binary when executed on target board crashes with:

root@OpenWrt:~# swanctl
Error relocating /usr/lib/ipsec/libstrongswan.so.0: XXX: symbol not found
Error relocating /usr/sbin/swanctl: lib: symbol not found

The binaries are compiled with gcc 8.4.0, binutils2.34, musl 1.1.24.
The unstripped binaries are ok and does not contain XXX symbol. When executed on target platform works fine.

However when I run `patchelf  --set-rpath '' libvici.so.0.0.0` or any other swanctl binary, then the unresolved XXX symbol appears in the library.

brenoguim:
In this PR I add a method to iterate over all fields of the ELF file (as best as we can) that represent indexes in strtable.

I use this method to check if the RPATH string being changed in `modifyRPATH` is shared among multiple fields. If it's shared, don't let patchelf taint the field.

We hope that patchelf continues to taint the vast majority of rpaths (as needed by Nix) but only refrain from doing so when it will break the binary.

Note: The first commit changes the section tainting from X to Z so I can look for the Xs in my tests.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1538,7 +1538,7 @@ void ElfFile<ElfFileParamNames>::modifyRPath(RPathOp op,
 
     /* !!! We assume that the virtual address in the DT_STRTAB entry
        of the dynamic section corresponds to the .dynstr section. */
-    auto shdrDynStr = findSectionHeader(".dynstr");
+    auto& shdrDynStr = findSectionHeader(".dynstr");
     char * strTab = (char *) fileContents->data() + rdi(shdrDynStr.sh_offset);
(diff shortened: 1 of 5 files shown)


Mic92:
bors merge

Issue #482, opened 2023-03-18
patchelf-0.17.2 seems to corrupt emacs on `staging-next`


trofi:
On current `staging-next` iteration quite a few emacs-dependent packages are failing. The failures seem to stem from the fact that `emacs` is incorrectly modified by `patchelf-0.17.2` (`0.15.0` works, bisected in `nixpkgs` by @mweinelt).

```
$ nix run https://github.com/NixOS/nixpkgs/archive/staging-next.tar.gz#emacs
Segmentation fault (core dumped)
```

```
$ patchelf --version
patchelf 0.17.2
```

It seems to have something to do with modified library list:

```
$ nix shell https://github.com/NixOS/nixpkgs/archive/staging-next.tar.gz#emacs
$ gdb emacs
Reading symbols from emacs...

warning: Loadable section ".dynstr" outside of ELF segments
  in /nix/store/lzahvwakhghr8b3ri40s935bwhn7nf0x-emacs-28.2/bin/emacs-28.2

warning: Loadable section ".dynamic" outside of ELF segments
  in /nix/store/lzahvwakhghr8b3ri40s935bwhn7nf0x-emacs-28.2/bin/emacs-28.2
(No debugging symbols found in emacs)
(gdb) run
Starting program: /nix/store/lzahvwakhghr8b3ri40s935bwhn7nf0x-emacs-28.2/bin/emacs

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7fe6597 in dl_main ()
   from /nix/store/8xk4yl1r3n6kbyn05qhan7nbag7npymx-glibc-2.35-224/lib/ld-linux-x86-64.so.2
(gdb) bt
#0  0x00007ffff7fe6597 in dl_main ()
   from /nix/store/8xk4yl1r3n6kbyn05qhan7nbag7npymx-glibc-2.35-224/lib/ld-linux-x86-64.so.2
#1  0x00007ffff7fe2a06 in _dl_sysdep_start ()
   from /nix/store/8xk4yl1r3n6kbyn05qhan7nbag7npymx-glibc-2.35-224/lib/ld-linux-x86-64.so.2
#2  0x00007ffff7fe45ad in _dl_start ()
   from /nix/store/8xk4yl1r3n6kbyn05qhan7nbag7npymx-glibc-2.35-224/lib/ld-linux-x86-64.so.2
#3  0x00007ffff7fe33a8 in _start ()
[...]

PR #485, opened 2023-03-20
Resize segment mapping rewritten sections if needed #482

brenoguim:
In some cases, the first segment in the executable doesn't map all content up to the next segment. So, sometimes there will be space in the beginning of the file to rewrite sections, but the first segment won't map the whole thing.
In this case, we extend the that segment to include all the needed space.

For this test I had to zero all the progbits and zip the binary. It went from 6MB to 166kb.


tests/short-first-segment.sh
```diff
@@ -0,0 +1,24 @@
+#! /bin/sh -e
+
+PATCHELF=$(readlink -f "../src/patchelf")
+SCRATCH="scratch/$(basename "$0" .sh)"
+READELF=${READELF:-readelf}
+
+EXEC_NAME="short-first-segment"
+
+if test "$(uname -i)" = x86_64 && test "$(uname)" = Linux; then
```
Mic92, line 9: Maybe the test should not fail if the package is not installed. gzip should be more mainstream on elf-based platforms than zip.

```suggestion
if test "$(uname -i)" = x86_64 && test "$(uname)" = Linux && gzip --version 2>/dev/null; then
```

Mic92:
@brenoguim thanks for taking over this one. I did look at ldc in the past but not long enough to understanding the issue.

tests/short-first-segment.sh
Mic92: Fixed.

Mic92:
bors merge

PR #487, opened 2023-03-23
add --update-rpath to change between RPATH and RUNPATH

adamse:
Thank you!

### DIFF
--- a/patchelf.1
+++ b/patchelf.1
@@ -79,6 +79,10 @@ the "/foo/lib" reference instead of the "/tmp" entry.
 .IP --print-rpath
 Prints the DT_RUNPATH or DT_RPATH for an executable or library.
 
+.IP --update-rpath
+Updates the DR_RPATH to a DT_RUNTPATH or vice versa if "--force-rpath" is given
+too.
+
 .IP --force-rpath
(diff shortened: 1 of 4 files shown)


patchelf.1
```diff
@@ -79,6 +79,10 @@ the "/foo/lib" reference instead of the "/tmp" entry.
 .IP --print-rpath
 Prints the DT_RUNPATH or DT_RPATH for an executable or library.
 
+.IP --update-rpath
+Updates the DR_RPATH to a DT_RUNTPATH or vice versa if "--force-rpath" is given
```
Mic92, line 83: Can you elaborate on when you want to set the one over the other?

adamse: They have slightly different semantics as detailed in `man ld-linux`. Would saying that + linking to the other manpage be ok? Or should I add a small summary?

Mic92: I think this get a different name. It's already confusing that we say --print-rpath and it's actually printing the DT_RUNPATH. "Updates" might be also the wrong verb as it convert from DR_RPATH to a DT_RUNPATH, whereas updates clashes already with `--set-rpath`

PR #488, opened 2023-03-26
Update build-id.sh

cedric84:
This prevents the test to fail when readelf is localized.

Thank you!


brenoguim:
Yey! Thanks for this. Being from Brazil, this test always failed for me too.

brenoguim (approved):

PR #490, opened 2023-04-08
Fix #310, add zsh completion

### LINKED ISSUE #310: Add zsh completion support
I found that patchelf lacks shell completion support.
I am willing to write zsh completion scripts for patchelf. If it’s acceptable, I’m happy to send a Pull Request

Freed-Wu:
Current commit can work. However, there are many same information in `--help`,
manpage and shell completions, so I think if we can generate them from one data
struct may be better.


Mic92:
bors merge

Mic92:
Thanks!

PR #494, opened 2023-04-29
chore(CI): run tests on CentOS 7

mayeut:
Towards getting reproducers/non-regression tests for #492 which is OS dependent.


.github/workflows/ci.yml
```diff
@@ -28,3 +28,28 @@ jobs:
...
+        c++ --version
+        ld --version
+        autoconf --version
+        ./bootstrap.sh
+        mkdir build && cd build
+        ../configure
+        make -j$(nproc) check || (cat tests/test-suite.log; exit 1)
+        EOF
+
+        docker run -v $(pwd):/gha quay.io/pypa/manylinux2014_${{ matrix.platform }}:latest sh -ec "cd /gha && sh ./build.sh"
```
Mic92, line 55: What would we need to upgrade to next year when centos 7 is EOL? https://wiki.centos.org/About/Product

mayeut: I think either ubuntu 18.04 or AlmaLinux 8 (or manylinux_2_28 depending on tooling available on AlmaLinux8).

PR #495, opened 2023-05-04
Fix wrong completion

Freed-Wu:
Sorry for a typo. According to <https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Completion-Functions>'s `_arguments`, `--long-option[description of option]:name of argument:completion method` should have 2 `:`s, not 3. Anyone can use `patchelf --set-interpreter <TAB>` to trigger this bug. And `patchelf --print-*` should be followed by a elf name: print some information about this elf then exit. So it shouldn't complete any other option `(-)`, but should complete elf name (not `(- : *)`, `(*)` will stop complete elf name).


Freed-Wu:
@Mic92

Freed-Wu:
@Mic92 @domenkozar

Freed-Wu:
@Mic92 @domenkozar @brenoguim

Freed-Wu:
@Mic92 @domenkozar @brenoguim Any problem?

Freed-Wu:
@edolstra @Mic92 @domenkozar @brenoguim Any problem?

Mic92:
No just lack of time. Things will get better next month.

Mic92:
Haven't compared the two but we also have now https://github.com/NixOS/patchelf/pull/552/files

Please re-open if there are still things missing in the other PR that this one fixes.

PR #499, opened 2023-06-02
Update elf.h based on Glibc

loongson-zn:
Update elf.h based on [Glibc](https://sourceware.org/git/glibc.git).


Mic92:
bors merge

PR #512, opened 2023-07-31
Do nothing if interpreter has not changed

JPEWdev:
If there is not change to the interpreter, don't make any modifications to the file. This is more efficient than requiring programs to check before changing, and also prevents thrashing of the ELF file, since changing the interpreter repeatedly will keep growing the ELF file each time, even if it's the same value.

This is particularly helpful where the same binary is symlinked or hardlinked with multiple names and has patchelf run on it.

Thank you!


shr-project:
Duplicates https://github.com/NixOS/patchelf/pull/508

JPEWdev:
So it does

Issue #520, opened 2023-09-20
patchelf breaks dylibs from recent Firefox Nightly builds


K900:
**Describe the bug**

After using `patchelf --set-rpath` on a library from a recent Firefox Nightly build, the library can no longer be loaded because it segfaults the linker.

**Steps To Reproduce**

- Download and unpack https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-central/firefox-119.0a1.en-US.linux-x86_64.tar.bz2
- Attempt to dlopen any of the .so files (`libmozsqlite3.so` was my test target), e.g. with `python -c 'import ctypes; ctypes.cdll.LoadLibrary("./libmozsqlite3.so")'
- Observe success
- `patchelf --set-rpath "test" ./libmozsqlite3.so`
- Attempt `dlopen` again
- Segfault

**Expected behavior**

No segfault.

**`patchelf --version` output**

Attempted both default nixpkgs 0.15.0 and current nixpkgs `patchelfUnstable` (c40128936fc804b74abddaa0bc1d8ef6e5dba48e).

**Additional context**

This seems to have been caused by upstream enabling some kind of advanced linker wizardry called "relrhack": https://hg.mozilla.org/mozilla-central/rev/032b87ff55061bcbdc7a85d9e18fde814797073a

The last build before that commit works fine.

glandium:
The problem is that the source code uses the _DYNAMIC symbol, which translates to the binary code accessing the .dynamic section at a fixed address. But patchelf moves it, and puts something else where it used to be, so the code reads garbage.

K900:
Well this is fun. So I guess we need a custom fixup for this...

glandium:
Smaller (independent) reproducer:
```
#include <stdio.h>
#include <elf.h>

extern __attribute__((visibility("hidden"))) Elf64_Dyn _DYNAMIC[];

int main() {
	for (Elf64_Dyn* dyn = _DYNAMIC; dyn->d_tag != DT_NULL; dyn++) {
		printf("%lx %p\n", dyn->d_tag, dyn->d_un.d_ptr);
	}
	return 0;
}
```
- compile with `gcc -o test test.c`
- run `./test`
- `patchelf --set-path foo test`
- run again. It will show garbage.

K900:
Yeah sounds like we just have to special case that symbol. Not that it's not already special cased by the linker...

glandium:
[...]

PR #521, opened 2023-09-21
feat: add --no-clobber-old-sections switch

K900:
Works around #520, may be useful for other cursed self-modifying things.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -58,6 +58,7 @@
 static bool debugMode = false;
 
 static bool forceRPath = false;
+static bool clobberOldSections = true;
 
 static std::vector<std::string> fileNames;
 static std::string outputFileName;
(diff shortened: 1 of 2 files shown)


src/patchelf.cc
```diff
@@ -664,14 +665,16 @@ template<ElfFileParams>
...
-       clobbering previously written new section contents. */
-    for (auto & i : replacedSections) {
-        const std::string & sectionName = i.first;
-        const Elf_Shdr & shdr = findSectionHeader(sectionName);
-        if (rdi(shdr.sh_type) != SHT_NOBITS)
-            memset(fileContents->data() + rdi(shdr.sh_offset), 'Z', rdi(shdr.sh_size));
+    if (clobberOldSections) {
+        /* Overwrite the old section contents with 'Z's.  Do this
+        *before* writing the new section contents (below) to prevent
+        clobbering previously written new section contents. */
```
vcunat, line 671: Nit: didn't indentation of the comment get broken?  (it changed and personally I don't like it much)

K900: Fixed.

src/patchelf.cc
```diff
@@ -2505,6 +2508,7 @@ static void showHelp(const std::string & progName)
   [--clear-execstack]\n\
   [--set-execstack]\n\
   [--rename-dynamic-symbols NAME_MAP_FILE]\tRenames dynamic symbols. The map file should contain two symbols (old_name new_name) per line\n\
+  [--no-clobber-old-sections]\t\tDo not clobber old section values - only use when executable does self-referential tricks.\n\
```
Mic92, line 2511: What are self-referential tricks? Can you add more details here?

K900: Basically anything that relies on things remaining at the same offset in the binary.

Mic92, line 2511:
```suggestion
  [--no-clobber-old-sections]\t\tDo not clobber old section values - only use when executables rely on sections remain at the same offset in the binary.\n\
```

K900:
How's this phrasing?

Mic92:
Better. Could you also update https://github.com/NixOS/patchelf/blob/master/patchelf.1 ?

K900:
Done. Tried to elaborate a bit on the behavior in the man page too.

PR #525, opened 2023-10-18
tests/short-first-segment: fix uname(1) usage and skip message

klemensn:
Replace non-portable GNU `-i, --hardware-platform` with `-m[, --machine]`
to fix "unknown error" usage error on, e.g. OpenBSD.

Also fix the check's backwards logic.  OpenBSD/spar64 now prints
	-skipping test: not supported on x86_64 Linux
	+skipping test: amd64 Linux required


klemensn:
Still relevant.

PR #529, opened 2023-11-05
Fix page size on Alpha

matoro:
All tests pass.

Also explicitly specifies -no-pie for executables which should have it disabled, to be compatible with gccs built with --enable-default-pie.


Mic92:
I had to partially revert this https://github.com/NixOS/patchelf/pull/541/commits/a9d3f085ffdc1730c722998ef95a632e2f76dc8f
Because it was breaking tests on other architectures.

matoro:
> I had to partially revert this [a9d3f08](https://github.com/NixOS/patchelf/commit/a9d3f085ffdc1730c722998ef95a632e2f76dc8f) Because it was breaking tests on other architectures.

Can you elaborate on this?  The reason I added that in is because in Gentoo, we build our gcc with `--enable-default-pie`, which means that PIE is on by default, and those executables needed to be built without PIE for them to work.

Mic92:
We had several errors like this: 

```
current interpreter is /lib/ld-musl-aarch64.so.1
running with explicit interpreter...
/lib/ld-musl-aarch64.so.1: ./simple: Not a valid dynamic program
FAIL set-interpreter-long.sh (exit status: 1)
```

Sorry. Didn't had the time to investigate what the issue is:

Here is the full run:

https://github.com/NixOS/patchelf/actions/runs/11890175871/job/33128295274

You should be able to reproduce this in CI by simply reverting my revert.

matoro:
> We had several errors like this:
> 
> ```
> current interpreter is /lib/ld-musl-aarch64.so.1
> running with explicit interpreter...
> /lib/ld-musl-aarch64.so.1: ./simple: Not a valid dynamic program
> FAIL set-interpreter-long.sh (exit status: 1)
> ```
> 
> Sorry. Didn't had the time to investigate what the issue is:
> 
> Here is the full run:
> 
> https://github.com/NixOS/patchelf/actions/runs/11890175871/job/33128295274
> 
> You should be able to reproduce this in CI by simply reverting my revert.

I just repro's this locally, it's because you're passing `-static` in `CXXFLAGS` to the test binaries as well.  This needs a configure flag so that it only gets passed to the `patchelf` binary unless the user specifies it.  Let me put up a PR to fix this.

matoro:
Never mind, I misread this.  I was able to reproduce it once on aarch64 musl, but it seems to have disappeared and it's back to passing.  Trying to figure out what happened.

Issue #531, opened 2023-11-05
`patchefd --set-rpath ... nodejs` broken since ~0.17.2 (0.15.0 works)


kvtb:
**Describe the bug**

`patchelf --set-rpath ${stdenv.cc.cc.lib}/lib node` produces broken executable with `patchelf-0.17.2` and `patchelf-0.18.0`, no problem with `patchelf-0.15.0`

the resulting executables crash with Segmentation Failed

`ldd node` produces no output at all

**Steps To Reproduce**

`node` is from this archive: https://nodejs.org/dist/v16.14.2/node-v16.14.2-linux-x64.tar.gz

Then run `patchelf --set-rpath ${stdenv.cc.cc.lib}/lib node` where `${stdenv.cc.cc.lib}` is the path to gcc libs path on your system

**Expected behavior**

Expected working executable as it was with `patchelf-0.15.0`

**`patchelf --version` output**

0.17.2, 0.18.0

**Additional context**

Downloaded `node` may seems ridiculous example, but using downloaded `node` and `protoc` it is an usual part of Maven build process.
BTW, there was no problem with patching `protoc` with any `patchelf` version

Patryk27:
Same as https://github.com/NixOS/patchelf/issues/244, fixed by https://github.com/NixOS/patchelf/pull/264 and (hopefully) will be fixed together with my recent pcloud-related things (since pcloud also relies on node under the hood).

Patryk27:
Alright, got it!

The issue is that `libnode.so` happens to have `.rodata` placed at the _beginning_ of the file:

```
; readelf -S libnode.so 
There are 29 section headers, starting at offset 0x152bf70:      
                                                                                             
Section Headers:                                                                             
  [Nr] Name              Type             Address           Offset  
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0 
  [ 1] .rodata           PROGBITS         0000000000000240  00000240
       00000000003796a8  0000000000000000 AMS       0     0     32
/* ... */
```

... which patchelf, when growing the program header table, _moves_ somewhere else (which also changes its virtual memory address, aii aii!).

Relocating this section (at least without relocating all the other sections) is illegal, because the underlying assembly might use (and, in this case, _does_ use) RIP-relative addressing, which breaks when the sections are shuffled around in the memory like that.

[...]

Issue #75, opened 2016-02-01
patchelf apparently doesn't update _DYNAMIC symbol?


mitchblank:
I was trying to make a function that determined if a program had already the correct runpath set (don't ask...)  This should be possible by inspecting the `_DYNAMIC` symbol as show here and just looking at my own `d_tag == DT_RUNPATH` entry:
  http://stackoverflow.com/a/2846736

However, after running a binary through patchelf, it seems like the `_DYNAMIC` symbol points to garbage?  I haven't investigated more, but I suspect that it just isn't updating that symbol after replacing that ELF header section... if that's the case than probably some other programs that are trying to be clever with `_DYNAMIC`-introspection

mitchblank:
Here is an example of what I'm talking about on a trivial "Hello, World" executable.  Before:

```
$ objdump -x -j .dynamic a.out | tail

Sections:
Idx Name          Size      VMA               LMA               File off  Algn
 20 .dynamic      000001d0  0000000000600e28  0000000000600e28  00000e28  2**3
                  CONTENTS, ALLOC, LOAD, DATA
SYMBOL TABLE:
0000000000600e28 l    d  .dynamic    0000000000000000              .dynamic
0000000000600e28 l     O .dynamic    0000000000000000              _DYNAMIC
```

So this shows that the `.dynamic` section is at file offset 0xe28 and will end up mapped at 0x600e28.  In the symbol table there is both a debugging symbol called `.dynamic` and also an object symbol called `_DYNAMIC` that both point to this section.

Then if I run patchelf on this binary, we'll see this change:

```
$ patchelf --set-rpath '$ORIGIN/../lib' a.out

$ objdump -x -j .dynamic a.out | tail

Sections:
Idx Name          Size      VMA               LMA               File off  Algn
  0 .dynamic      000001e0  00000000003ff270  00000000003ff270  00000270  2**3
                  CONTENTS, ALLOC, LOAD, DATA
SYMBOL TABLE:
00000000003ff270 l    d  .dynamic	0000000000000000              .dynamic
0000000000600e28 l     O .dynamic	0000000000000000              _DYNAMIC
```

So the section has now been relocated, both in the file and in virtual address space.  The debugging symbol `.dynamic` is also updated.  However the object symbol `_DYNAMIC` still has its old value and isn't pointing into this ELF table anymore.

mitchblank:
[...]

PR #264, opened 2021-02-19
Expand LOAD segment to cover .dynstr and .dynsym sections

ehmry:
Hopefully this fixes #244.

PR #544, opened 2024-02-06
Allocate PHT & SHT at the end of the *.elf file

Patryk27:
Closes https://github.com/NixOS/patchelf/issues/531.
Closes https://github.com/NixOS/patchelf/issues/482.
Closes https://github.com/NixOS/patchelf/issues/244.

Upstream-wise, affects https://github.com/NixOS/nixpkgs/issues/226339.
(didn't want to write `close` so that merging this merge request doesn't close that issue at once)

# Abstract

Patching an *.elf file might require extending its program header table, which can then cause it to run out of its originally allocated space (both in terms of file offset and virtual memory).

Currently patchelf solves this problem by finding which sections would overlap PHT and then by moving them to the end of *.elf file:

https://github.com/NixOS/patchelf/blob/7c2f768bf9601268a4e71c2ebe91e2011918a70f/src/patchelf.cc#L832

As compared to similar logic for binaries:

https://github.com/NixOS/patchelf/blob/7c2f768bf9601268a4e71c2ebe91e2011918a70f/src/patchelf.cc#L964

... the logic for libraries is missing a crucial check: it doesn't take into account whether that particular section _can_ be shuffled around - in particular, sections with `SHT_PROGBITS` can't!

As luck would have it, `libnode.so` (e.g. shipped together with `pcloud`) does have `.rodata` (a section with `SHT_PROGBITS` active) right at the beginning of the file:

```
; readelf -S libnode.so

There are 29 section headers, starting at offset 0x152bf70:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .rodata           PROGBITS         0000000000000240  00000240
       00000000003796a8  0000000000000000 AMS       0     0     32
/* ... */
```

... which patchelf happily moves around, breaking RIP-relative addressing in the assembly (which, after patching, tries to access the `ZZZZ`-ed memory).

This commit fixes the issue by changing the logic from:

> if there's no space for PHT, shuffle sections around

... to, perhaps a bit more wasteful in terms of storage:

> just always allocate PHT at the end of the file

As far as I've checked, the reason why PHT was so strictly kept at the beginning was an old Linux bug:

https://github.com/NixOS/patchelf/blob/7c2f768bf9601268a4e71c2ebe91e2011918a70f/src/patchelf.cc#L857
https://github.com/NixOS/patchelf/blob/7c2f768bf9601268a4e71c2ebe91e2011918a70f/BUGS#L1

... which is not present anymore (not sure when precisely was it fixed, though - the original entry in the BUGS file is dated 2005).

Seizing the day, I'm also including another fix (for binaries), so that merging this pull request will solve all pcloud-related problems.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -555,8 +555,7 @@ void ElfFile<ElfFileParamNames>::shiftFile(unsigned int extraPages, size_t start
 
     assert(splitIndex != -1);
 
-    /* Add a segment that maps the new program/section headers and
-       PT_INTERP segment into memory.  Otherwise glibc will choke. */
+    /* Add another PT_LOAD segment loading the data we've split above. */
     phdrs.resize(rdi(hdr()->e_phnum) + 1);
     wri(hdr()->e_phnum, rdi(hdr()->e_phnum) + 1);
     Elf_Phdr & phdr = phdrs.at(rdi(hdr()->e_phnum) - 1);
@@ -636,9 +635,17 @@ unsigned int ElfFile<ElfFileParamNames>::getSectionIndex(const SectionName & sec
 }
 
 template<ElfFileParams>
-bool ElfFile<ElfFileParamNames>::haveReplacedSection(const SectionName & sectionName) const
+bool ElfFile<ElfFileParamNames>::hasReplacedSection(const SectionName & sectionName) const
 {
     return replacedSections.count(sectionName);
+}
+
+template<ElfFileParams>
+bool ElfFile<ElfFileParamNames>::canReplaceSection(const SectionName & sectionName) const
+{
+    auto shdr = findSectionHeader(sectionName);
+
+    return sectionName == ".interp" || rdi(shdr.sh_type) != SHT_PROGBITS;
 }
 
 template<ElfFileParams>
@@ -823,28 +830,60 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
     unsigned int num_notes = std::count_if(shdrs.begin(), shdrs.end(),
         [this](Elf_Shdr shdr) { return rdi(shdr.sh_type) == SHT_NOTE; });
 
-    /* Because we're adding a new section header, we're necessarily increasing
-       the size of the program header table.  This can cause the first section
-       to overlap the program header table in memory; we need to shift the first
-       few segments to someplace else. */
-    /* Some sections may already be replaced so account for that */
+    /* Compute the total space needed for the replaced sections, pessimistically
+       assuming we're going to need one more to account for new PT_LOAD covering
+       relocated PHDR */
+    off_t phtSize = roundUp((phdrs.size() + num_notes + 1) * sizeof(Elf_Phdr) + sizeof(Elf_Ehdr), sectionAlignment);
+    off_t shtSize = roundUp(rdi(hdr()->e_shnum) * rdi(hdr()->e_shentsize), sectionAlignment);
+
+    /* Check if we can keep PHT at the beginning of the file.
+
+       We'd like to do that, because it preverves compatibility with older
+       kernels¹ - but if the PHT has grown too much, we have no other option but
+       to move it at the end of the file.
+
+       ¹ older kernels had a bug that prevented them from loading ELFs with
+         PHDRs not located at the beginning of the file; it was fixed over
+         0da1d5002745cdc721bc018b582a8a9704d56c42 (2022-03-02) */
+    bool relocatePht = false;
     unsigned int i = 1;
-    Elf_Addr pht_size = sizeof(Elf_Ehdr) + (phdrs.size() + num_notes + 1)*sizeof(Elf_Phdr);
-    while( i < rdi(hdr()->e_shnum) && rdi(shdrs.at(i).sh_offset) <= pht_size ) {
-        if (not haveReplacedSection(getSectionName(shdrs.at(i))))
-            replaceSection(getSectionName(shdrs.at(i)), rdi(shdrs.at(i).sh_size));
+
+    while (i < rdi(hdr()->e_shnum) && ((off_t) rdi(shdrs.at(i).sh_offset)) <= phtSize) {
+        const auto & sectionName = getSectionName(shdrs.at(i));
+
+        if (!hasReplacedSection(sectionName) && !canReplaceSection(sectionName)) {
+            relocatePht = true;
+            break;
+        }
+
         i++;
     }
-    bool moveHeaderTableToTheEnd = rdi(hdr()->e_shoff) < pht_size;
 
-    /* Compute the total space needed for the replaced sections */
-    off_t neededSpace = 0;
+    if (!relocatePht) {
+        unsigned int i = 1;
+
+        while (i < rdi(hdr()->e_shnum) && ((off_t) rdi(shdrs.at(i).sh_offset)) <= phtSize) {
+            const auto & sectionName = getSectionName(shdrs.at(i));
+            const auto sectionSize = rdi(shdrs.at(i).sh_size);
+
+            if (!hasReplacedSection(sectionName)) {
+                replaceSection(sectionName, sectionSize);
+            }
+
+            i++;
+        }
+    }
+
+    /* Calculate how much space we'll need. */
+    off_t neededSpace = shtSize;
+
+    if (relocatePht) {
+        neededSpace += phtSize;
+    }
+
     for (auto & s : replacedSections)
         neededSpace += roundUp(s.second.size(), sectionAlignment);
 
-    off_t headerTableSpace = roundUp(rdi(hdr()->e_shnum) * rdi(hdr()->e_shentsize), sectionAlignment);
-    if (moveHeaderTableToTheEnd)
-        neededSpace += headerTableSpace;
     debug("needed space is %d\n", neededSpace);
 
     Elf_Off startOffset = roundUp(fileContents->size(), alignStartPage);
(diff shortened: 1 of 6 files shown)


src/patchelf.cc
```diff
@@ -787,10 +801,51 @@ void ElfFile<ElfFileParamNames>::writeReplacedSections(Elf_Off & curOff,
             }
         }
 
-        curOff += roundUp(i->second.size(), sectionAlignment);
+        curOff += roundUp(i->second.data.size(), sectionAlignment);
     }
 
-    replacedSections.clear();
+    /* Expand the LOAD segment as neccessary */
```
Patryk27, line 807: This is the extra mandatory fix -- in fact, I think that's a regression from v0.15, originally posted here:

https://github.com/NixOS/patchelf/pull/264

Without this fix, `libnode.so` is broken to the point where it crashes inside the linker :smile:

src/patchelf.cc
```diff
@@ -1274,26 +1312,109 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
...
+
+                if (rs == replacedSections.end()) {
+                    // Entry refers to a section that didn't got moved around
+                    // (we still might have to update shndx, though)
+                } else {
+                    if (rdi(sym->st_value) < rs->second.prev_addr) {
+                        warn("entry %d refers to an address outside the section, skipping\n", entry);
+                        continue;
+                    } else {
+                        newValue = rdi(sym->st_value) - rs->second.prev_addr + rdi(shdrs.at(newIndex).sh_addr);
```
Patryk27, line 1359: This is part of the extra fix, for https://github.com/NixOS/patchelf/issues/75 - it doesn't seem to be required for the pcloud-nodejs problem, but I coded it before I stumbled upon the PHT thingie, that's why it's here.

Could be nice to keep, can be refactored into a separate MR as well, if review proves difficult here.

src/patchelf.cc
```diff
@@ -1274,26 +1312,109 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
...
+                (size_t) newValue
+            );
+
+            wri(sym->st_shndx, newIndex);
+            wri(sym->st_value, newValue);
+        }
+    }
+
+    /* Rewrite the .rela.dyn section */
+    for (unsigned int i = 1; i < rdi(hdr()->e_shnum); ++i) {
```
Patryk27, line 1388: ditto (second part of the extra fix)

tests/no-rpath-pie-powerpc.sh
```diff
@@ -37,9 +37,9 @@ if [ "$(echo "$readelfData" | grep -c "PHDR")" != 1 ]; then
 fi
 
 virtAddr=$(echo "$readelfData" | grep "PHDR" | awk '{print $3}')
-if [ "$virtAddr" != "0x00000034" ]; then
+if [ "$virtAddr" != "0x01040000" ]; then
```
Patryk27, line 40: Previously PHDR was always kept right after EHDR, while now it's allocated at the end of the file.

tests/repeated-updates.sh
```diff
@@ -34,7 +34,7 @@ load_segments_after=$(${READELF} -W -l libbar.so | grep -c LOAD)
 # To be even more strict, check that we don't add too many extra LOAD entries
 ###############################################################################
 echo "Segments before: ${load_segments_before} and after: ${load_segments_after}"
-if [ "${load_segments_after}" -gt $((load_segments_before + 2)) ]
+if [ "${load_segments_after}" -gt $((load_segments_before + 3)) ]
```
Patryk27, line 37: _I think_ that's because now we might allocate a brand-new LOAD segment just for the PHDR's sake, but I haven't checked it thoroughly yet.

src/patchelf.cc
```diff
@@ -1274,26 +1312,109 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
...
+
+        if (rdi(shdr.sh_type) != SHT_RELA) {
+            continue;
+        }
+
+        debug("rewriting relocations (section %d)\n", i);
+
+        unsigned int r_idx = 0;
+
+        for (auto& r: getSectionSpan<Elf64_Rela>(shdr)) {
```
Patryk27, line 1399: TODO: don't hardcode Elf64_Rela here

src/patchelf.cc
```diff
@@ -1274,26 +1312,109 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
...
+            if (ELF64_R_TYPE(r.r_info) == R_X86_64_RELATIVE) {
```
Patryk27, line 1400: aarch64 relatives (and possibly others) have the same format, maybe we could support them as well?

src/patchelf.cc
```diff
@@ -1274,26 +1312,109 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
...
-                if (shndx >= sectionsByOldIndex.size()) {
-                    fprintf(stderr, "warning: entry %d in symbol table refers to a non-existent section, skipping\n", shndx);
-                    continue;
+
+            if (shndx == SHN_UNDEF || shndx >= SHN_LORESERVE) {
+                continue;
+            }
+
+            if (shndx >= sectionsByOldIndex.size()) {
+                warn(
```
Patryk27, line 1331: TODO: wrong formatting

src/patchelf.cc
```diff
@@ -1033,31 +1022,33 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsExecutable()
...
-    /* Clear out the free space. */
-    Elf_Off curOff = sizeof(Elf_Ehdr) + phdrs.size() * sizeof(Elf_Phdr);
-    debug("clearing first %d bytes\n", startOffset - curOff);
-    memset(fileContents->data() + curOff, 0, startOffset - curOff);
+    /* Ensure PHDR is covered by a LOAD segment.
 
+       Because PHDR is supposed to have been covered by such section before, in
+       here we assume that we don't have to create any new section, but rather
+       extend the existing one. */
+    for (auto& phdr : phdrs)
```
Patryk27, line 1034: This logic used to be run only if `neededSpace <= startOffset`, under the assumption that otherwise (when `neededSpace > startOffset`), we call `shiftFile()`, which will automatically create appropriate `PT_LOAD` segment:

https://github.com/NixOS/patchelf/blob/7c2f768bf9601268a4e71c2ebe91e2011918a70f/src/patchelf.cc#L557

But _that_ was done under the assumption that `shiftFile()`'s `startOffset` will land inside a loadable segment that happens to contain `PHDR`, which is not necessarily true!  Some binaries have a dedicated small `LOAD` segment just for `PHDR`, making `shiftFile()` then pick the next available `LOAD` segment.

I believe this is the same case this merge request tried to cover (or at least both make `pcloud` a valid binary):

https://github.com/NixOS/patchelf/pull/264

This is now tested as a part of the `short-first-segment` test.

Patryk27:
Btw, I've got another merge request in the works, one that fixes https://github.com/NixOS/patchelf/issues/75 and hopefully will be able to handle https://github.com/NixOS/patchelf/issues/520 as well, but I've ultimately decided to move it out of this pull request, since it's technically a separate changeset (and less important, at least from my point of view) -- I'll prepare it in the upcoming weeks.

src/patchelf.cc
Patryk27: Hmm, in a hindsight, now I'm thinking whether this approach can't cause some segments to overlap each other? (maybe that's why tests on a few other architectures fail, as I now see)

jvolkman:
Another issue is that `strip` doesn't like binaries with with the PHT and SHT bits at the end, and will either fail or generate invalid binaries as output. One workaround is to just strip first, but I guess this is not possible in all workflows. 

See #117 and #10.

Definitely seems like a `strip`/`libbfd` bug since, as far as can tell, nothing in the ELF spec requires these tables to be at the top. I spent a couple days stepping through [bfd/elf.c](https://github.com/bminor/binutils-gdb/blob/master/bfd/elf.c) back when I was working on [this](https://github.com/jvolkman/repairwheel/blob/main/src/repairwheel/linux/elffile.py) (which also places the updated headers at the end), but I was ultimately unable to decipher what was going on.

tests/short-first-segment.sh
```diff
@@ -24,8 +24,10 @@ cd "${SCRATCH}"
 
 ldd "${EXEC_NAME}"
 
-${PATCHELF} --add-rpath lalalalalalalala --output modified1 "${EXEC_NAME}"
+${PATCHELF} --set-rpath "$(tr -dc A-Za-z0-9 </dev/urandom | head -c 4096)" --output modified1 "${EXEC_NAME}"
```
Mic92, line 27: I would make the test reproducible and not have a random value here i.e. storing a static secret somewhere.

Mic92:
@Patryk27 rebased CI and it looks like this causes several regressions on non-x86 platforms.

Mic92:
It should be possible to reproduce this locally using `boot.binfmt.emulatedSystems` and than either with a cross compiler from nixpkgs or with the docker image used in the tests.

Patryk27:
Nope, still fails on other architectures (I'm testing using binfmt) - currently hitting `Inconsistency detected by ld.so: rtld.c: 1280: rtld_setup_main_map: Assertion `GL(dl_rtld_map).l_libname' failed!` on some tests. Analyzing further 🔍

Mic92:
If we can skip tests if they run in qemu emulation, that would be nice. Patchelf is full of platform assumption and we would degraded it very quickly without proper testing. I will try to add some native arm64 ci here, so we can test that part. We should cover at least amd64/arm64.

Patryk27:
Alright, @Mic92 - v3.0 is ready!

I've adjusted the logic so that we keep PHDR at the beginning of the file for as long as it's possible - if we detect that the new PHDR would overflow another section, only then we relocate it at the end of the file.

This makes the ELFs generated by patchelf compatible with both the older and newer kernels (unless we have to launch the new logic, e.g. for the pCloud's NodeJS library - those ELFs would be incompatible with the older kernels; but there's no other solution here).

I still need to write an extra test, though - for now I've just manually checked that the pCloud launches when patchelf-ed with my approach here.

Mic92:
> Inconsistency detected by ld.so: dl-setup_hash.c: 36: _dl_setup_hash: Assertion `(bitmask_nwords & (bitmask_nwords - 1)) == 0' failed!

This from the test suite on x86_64.

Patryk27:
Huh, interesting - all tests pass on my machine. I’ll take a look.

Mic92:
@Patryk27 And you say that cherry-picking the qemu patch, would fix the other tests?

osandov:
> I've adjusted the logic so that we keep PHDR at the beginning of the file for as long as it's possible - if we detect that the new PHDR would overflow another section, only then we relocate it at the end of the file.

Drive by comment here, thanks for making this change! There's another important reason to try to keep the PHDRs at the beginning of the file: by default, Linux only saves the first page of an executable or shared library in core dumps, and debuggers need the PHDRs and build ID from a core dump to identify the exact binary that was in use. So if the PHDRs aren't in the first page, debuggers won't be able to find what they need.

That was the motivation for the binutils patch (https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=bf6d7087de0a7351fd1dfd5f41522a7f4f576180) that caused #568 and NixOS/nixpkgs#362211, which I assume this also fixes?

Patryk27:
@Mic92 

> And you say that cherry-picking the qemu patch, would fix the other tests? 

With my current approach the cherry-pick is not required, since we generate ELFs compatible with both kernels (and after I add a test covering the new behavior, I'll enable it just for x86_64, to avoid going through qemu-user).

Mic92:
@Patryk27 Ok, but why do all emulated tests fail than a couple of tests each.

Patryk27:
Okie, should be good now 🤞🏻

Patryk27:
Ha, seems I got it this time! cc @Mic92

hroncok:
FWIW, when we applied this patch in Fedora, the maintainer said:

> I've applied it, but a simple test with --set-rpath causes a segfault, unless built with -Wl,--no-rosegment

https://src.fedoraproject.org/rpms/patchelf/pull-request/5#comment-243635

cc @jeremysanders

jeremysanders:
[example.zip](https://github.com/user-attachments/files/18647900/example.zip)

Well, this is just a simple hello world, compiled on Fedora 41, using

    gcc -o mytest mytest.c

I ran

    patchelf --set-rpath /usr/lib mytest

Running then segfaults.

Patryk27:
Got it, I'll take a look tomorrow!

Patryk27:
Ok, this is a known problem, not related to this patch (i.e. older patchelf also generates an invalid binary):
- https://github.com/NixOS/patchelf/issues/568
- https://github.com/NixOS/nixpkgs/issues/362211

I see if this is fixable, but it's a parallel issue to what this patch was trying to fix.

Patryk27:
Well - this works, but I haven't checked if it makes much sense:
```patch
diff --git a/src/patchelf.cc b/src/patchelf.cc
index 14c0802..3d6f3bc 100644
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1008,7 +1008,7 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsExecutable()
         /* !!! Why do we stop after a .dynstr section? I can't
            remember! */
         if ((rdi(shdr.sh_type) == SHT_PROGBITS && sectionName != ".interp")
-            || prevSection == ".dynstr")
+            || prevSection == ".dynstr" || sectionName == ".note.gnu.build-id")
         {
             startOffset = rdi(shdr.sh_offset);
             startAddr = rdi(shdr.sh_addr);
```

Ericson2314:
@Patryk27 

> Btw, I've got another merge request in the works, one that fixes #75 and hopefully will be able to handle #520 as well, but I've ultimately decided to move it out of this pull request, since it's technically a separate changeset (and less important, at least from my point of view) -- I'll prepare it in the upcoming weeks.

Do you still have this? Do you want to put it up as a draft?

Ericson2314:
Thanks for the link.

> but if you think it will come handy, I could try to revive it; if you have a binary that's broken by this behavior being unimplemented, I could also hug it, it would make the testing easier.

So right now there are a bunch of issues in later releases, a bit of a "patches that fixed some bugs causes others bugs" situations. And I am trying to figure out more what's going on, than I am sure what to do about it.

So for now I would say don't try just yet, save your effort. But, it's very good the work wasn't lost. Probably in a short while, once I/other maintainers have a better sense of what's going on, we'll give you the go-ahead. And ideally at that point, we *do* know what the binaries we are trying to fix are.

Patryk27:
Sure, got it - waiting for further instructions / questions then 🙂

Issue #549, opened 2024-03-11
_arguments:463: command not found: dynamic


zhefox:
**Describe the bug**

➜  / patchelf --set-interpreter /root/tools/glibc_arguments:463: command not found: dynamic27.so ./chatting
_arguments:463: command not found: dynamic
_arguments:463: command not found: dynamic
patchelf                   /root/tools/glibc

**Steps To Reproduce**

install zsh and history plugin then tab to complement path

**`patchelf --version` output**

➜  / patchelf --version
patchelf 0.18.0

zhefox:
plugins=(
        git
        zsh-autosuggestions
        zsh-syntax-highlighting
)

RocketMaDev:
Check out my pr above :)
I fixed the problem you mentioned and enhanced it. Please check your fpath (check out by `echo $fpath`), download `_patchelf` in my fork, and replace your existed `_patchelf`, which is in one of the fpaths. Restart your zsh, and the correct completion will take effect.

zhefox:
thanks,i will try

zhefox:
I confirm you fixed it and I will close this bug issue

RocketMaDev:
@zhefox I pushed a new commit, now files like `./chatting` can get dep completion. Plz download my latest edition to get a better experience.

zhefox:
> @zhefox I pushed a new commit, now files like `./chatting` can get dep completion. Plz download my latest edition to get a better experience.

Thank u, I have updated it.

LittFlower:
> Check out my pr above :) I fixed the problem you mentioned and enhanced it. Please check your fpath (check out by `echo $fpath`), download `_patchelf` in my fork, and replace your existed `_patchelf`, which is in one of the fpaths. Restart your zsh, and the correct completion will take effect.

Sorry, I have tried the latest version of commit (which I compiled myself), but I still have problems with the completion in zsh. What should I do?

RocketMaDev:
@LittFlower please confirm that zsh run the right completion script.

LittFlower:
Ohhhh.... This is my negligence, and I apologize for it.... I'm sorry for wasting your time and thank you very much for your reminder!

PR #552, opened 2024-04-06
fix & feat: enhanced completion for zsh

RocketMaDev:
I fixed zsh completion which is not working properly (like #549), now files can be completed. What's more, I add a feature that when triggering `Tab` where dependencies should be entered, the dependencies of the elf will pop out. Here's a screenshot:

Typical case

If the elf is the first arg, it works, too:

File is in the front

In any other cases (file is not exist, or is a `.so`, or is not exist), the match fallback to the default _files match.

Error handling

I tried my best to test all the cases on my machine and it should be right.


RocketMaDev:
I've tested a lot and everything works. Would there be someone help me merge it in?

PR #566, opened 2024-09-05
Fix alignment problem when rewriting sections

pablogsal:
After commit ac212d0e6fb8b741e5a5e9ea61091149103f401c the code to
rewrite alignment section has been changed to use the largest alignment
in the list of segments instead of the alignment that it's retrieved
using getPageSize().

Unfortunately the code didn't update the offset as well to keep the
invariant p_vaddr % alignment == p_offset % alignment.

Thank you!


lkollar:
I think this fixes #492.

boegel:
Without this change, we're running into the "` ELF load command address/offset not properly aligned`" issue reported #492 .

Are there any updates on getting this merged and a new version of `patchelf` released?

PR #570, opened 2024-11-15
Add `--add-rpath-and-shrink` option to do both operations in one go

Ma27:
__Note:__ I know this is missing an update of the man-page and tests. I'd like to know if you'd be open to accept such a patch in the first place before investing more time into it.

The core issue is that `--shrink-rpath` doesn't free up space filled
with `X` by `--add-rpath`, so using a combination of `--add-rpath` &
`--shrink-rpath` on a virtual env will blow up shared libraries and
cause them to segfault when loaded by Python at some point.

This is not an issue with Nix builds, but it absolutely is when using this
in an imperatively managed virtualenv.

cc @Mic92 

---

Thank you!

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,3 +1,3 @@
-(import (fetchTarball https://github.com/edolstra/flake-compat/archive/master.tar.gz) {
+(import (fetchTarball "https://github.com/edolstra/flake-compat/archive/master.tar.gz") {
   src = ./.;
 }).defaultNix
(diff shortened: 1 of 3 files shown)


Mic92:
@Ma27 I can probably just give you merge access to this project.
I am currently not doing here much, because it feels a bit difficult to do any meaningful changes without breaking some random corner of nixpkgs.

Mic92:
Feel free to go ahead with this pull request: https://github.com/orgs/NixOS/teams/patchelf-maintainers/members

Ma27:
So first of all, your trust is greatly appreciated, thanks! ❤️ 

I'll try to bring this patch into a cleaner state within the next weeks to get it in.

That being said, I'm afraid I won't be able to exercise any further maintainer duties in a forseeable future (i.e. the next few months): I'm working on finishing my studies soonish and I can't afford a(nother) can of worms that a patchelf bump is for now (I know how painful updates in nixpkgs are).

If nothing has changed about the situation here in a few months, I'll reconsider my decision.

PR #575, opened 2024-11-19
Do not build test binaries statically in CI

matoro:
For the CI runs intended to be testing static binary, only build the main `patchelf` binary statically, and not the test binaries.  This allows us to restore the correct behavior for toolchain builds with PIE enabled by default.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -89,7 +89,7 @@ jobs:
           path: dist
       - name: Build binaries
         env:
-          CXXFLAGS: "-D_FORTIFY_SOURCE=2 -fstack-protector-strong -Wformat -Werror=format-security -O2 -static"
+          CXXFLAGS: "-D_FORTIFY_SOURCE=2 -fstack-protector-strong -Wformat -Werror=format-security -O2"
         run: |
           cat <<EOF > build.sh
           set -e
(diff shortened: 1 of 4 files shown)


Mic92:
Sounds ok to me in principle

matoro:
This is clearly not the right solution, as evidenced by it still failing and also upon closer inspection I misread this - `-static` is not being passed to the test binaries.

Despite running on aarch64 musl I am unable to reliably reproduce this.

Mic92:
You cannot reproduce this when pulling the same container locally?

matoro:
> You cannot reproduce this when pulling the same container locally?

I was running in a Gentoo aarch64 musl chroot, and yeah it doesn't reproduce there.

```
# uname -m
aarch64
# /lib/ld-musl-aarch64.so.1
musl libc (aarch64)
Version 1.2.4
Dynamic Program Loader
Usage: /lib/ld-musl-aarch64.so.1 [options] [--] pathname [args]
# git clone "https://github.com/nixos/patchelf"
...
# pushd patchelf
...
# git revert --no-edit d985189
...
# ./bootstrap.sh
...
# CXXFLAGS="-D_FORTIFY_SOURCE=2 -fstack-protector-strong -Wformat -Werror=format-security -O2 -static" ./confi
gure
...
# make check
...
============================================================================
Testsuite summary for patchelf 0.18.0
============================================================================
# TOTAL: 58
# PASS:  56
# SKIP:  2
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

PR #582, opened 2025-01-08
Revert "Add `riscv64` support for flake.nix"

Mic92:
This reverts commit f163e8df3d0fda67cd2e897097ed760fafd6b5f2.

We don't support riscv64 in hydra. This needs to be refactored to only enable riscv64 in the flake but not in hydra jobs.


Mic92:
@mergify queue

PR #589, opened 2025-02-14
fix: specify zsh completions in patchelf.spec.in

murkvin:
Problem: When building the patchelf RPM, zsh completions are installed (but not packaged). This causes an RPM build failure.

Solution: We need to identify the zsh completions in the generated RPM spec file.

Testing: I updated the RPM spec template file and successfully built the RPM.


Mic92:
@mergify queue

PR #595, opened 2025-04-21
Fix assertion failure when setting the interpreter of a debug-only executable

corngood:
The included test currently fails only on binutils 2.44, which is available on `nixpkgs-unstable`.  I've only tested on x86_64-linux.

Perhaps `lastReplaced` should be renamed because it no longer ends up pointing to the actual last replaced section.

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -997,9 +997,6 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsExecutable()
        Stop when we reach an irreplacable section (such as one of type
        SHT_PROGBITS).  These cannot be moved in virtual address space
        since that would invalidate absolute references to them. */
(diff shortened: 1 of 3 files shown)


corngood:
Should we also bail out of `setInterpreter` if .interp is NOBITS?  If so, should it be an error?

Unfortunately auto-patchelf calls `--set-interpreter` on any ELF where .interp exists, even if it's NOBITS, so if we make it an error, changes will still be required there.

corngood:
The test I added showed another problem when run with clang.  It looks like clang outputs ET_DYN, which means it doesn't even hit the other fix in `rewriteSectionsExecutable`.

The problem with clang (with both old and new binutils) is that it's accessing .dynamic even if it's NOBITS.  I've pushed a fix for that in ad3d3d4abbcff52e142732b04dd27257d35285b6.

PR #596, opened 2025-05-01
Fix --remove-needed not updating .gnu.version_r

### LINKED ISSUE #214: --remove-needed does not update .gnu.version_r table
**Describe the bug**
with the newest version of patchelf (pathelf-0.11 and master branch), remove `so` will cause a bug 
```
Inconsistency detected by ld.so: dl-version.c: 205: _dl_check_map_versions: Assertion `needed != NULL' failed!
```

**Steps To Reproduce**

deps: libonnxruntime.so https://drive.google.com/file/d/1rQZYIkzOu4UGJ4Cczm-FVM2LrxYw1Xc0/view?usp=sharing

1. remove libs
```
$ patchelf --remove-needed libcublas.so.10.0 libonnxruntime.so
```

2. read so
```
$ ldd libonnxruntime.so
Inconsistency detected by ld.so: dl-version.c: 205: _dl_check_map_versions: Assertion `needed != NULL' failed!
```

**other**
the problem here, looks like things in `replaceNeeded`, updating the `.gnu.version_r` section

any clue to solve this problem will be highly appreciated.

** one more thing**
I tried to modify the `removeNeeded` with the following:
```
unsigned int verNeedNum = 0;

    Elf_Dyn * dyn = (Elf_Dyn *) (contents + rdi(shdrDynamic.sh_offset));
    Elf_Dyn * last = dyn;
    for ( ; rdi(dyn->d_tag) != DT_NULL; dyn++) {
        if (rdi(dyn->d_tag) == DT_NEEDED) {
            char * name = strTab + rdi(dyn->d_un.d_val);
            if (libs.find(name) != libs.end()) {
                debug("removing DT_NEEDED entry '%s'\n", name);
                changed = true;
            } else {
                debug("keeping DT_NEEDED entry '%s'\n", name);
                *last++ = *dyn;
            }
        } else
            *last++ = *dyn;

        // this is what i add
        if (rdi(dyn->d_tag) == DT_VERNEEDNUM) {
            verNeedNum = rdi(dyn->d_un.d_val);
        }
    }

    memset(last, 0, sizeof(Elf_Dyn) * (dyn - last));

  if (verNeedNum) {
        auto shdrVersionR = findSection(".gnu.version_r");
        Elf_Shdr & shdrVersionRStrings = shdrs[rdi(shdrVersionR.sh_link)];
        char * verStrTab = (char *) contents + rdi(shdrVersionRStrings.sh_offset);
        std::string versionRStringsSName = getSectionName(shdrVersionRStrings);

        debug("found .gnu.version_r with %i entries, strings in %s\n", verNeedNum, versionRStringsSName.c_str());

        unsigned int verStrAddedBytes = 0;
        Elf_Verneed * need = (Elf_Verneed *) (contents + rdi(shdrVersionR.sh_offset));
        while (verNeedNum > 0) {
            char * file = verStrTab + rdi(need->vn_file);
            auto i = libs.find(file);
            if (i != libs.end()) {
                debug("removing .gnu.version_r entry '%s'\n", file);
                debug("resizing string section %s ...\n", versionRStringsSName.c_str());
                
               // THIS COULD BE WRONG
                wri(need->vn_file, rdi(shdrVersionRStrings.sh_size));

                changed = true;
            }
            // the Elf_Verneed structures form a linked list, so jump to next entry
            need = (Elf_Verneed *) (((char *) need) + rdi(need->vn_next));
            --verNeedNum;
        }
    }
```
this could led to another problem
```
$ ldd libonnxruntime.so
./libonnxruntime.so: ./libonnxruntime.so: version `libcublas.so.10.0' not found (required by ./libonnxruntime.so)
```

and when use `libonnxruntime.so` :
```
java.lang.UnsatisfiedLinkError: /tmp/onnxruntime-java3138521611266514151/libonnxruntime.so: java: version `libcublas.so.10.0' not found (required by /tmp/onnxruntime-java3138521611266514151/libonnxruntime.so)
```

outtersg:
Fixes #214

Constructed after reading @njsmith's great work on #84 / #85 (thus some variable names being the same).

#### Technical considerations

Even if the section is theorically a compact array of same-size Elf_Verneed entries (thus `sh_size / sizeof(ElfXX_Verneed)` should give the count of entries, which is written in 1 or 2 places too), there is redundancy with a chained list where each entry has the next entry's offset in its `vn_next` field (with a value of `0` ending the list… thus giving a third way to know we have reached the end, in addition with the count and the section size / entry size methods thereabove).

So there are possible holes in the section (when `vn_next > sizeof(ElfXX_Verneed)`), and maybe (admittedly stealthy) data to preserve in those holes.

To adapt to that uncertainty, the patch tries to be **as conservative as possible** in its writes:\
instead of shifting the last entries to occupy the place let by the removed ones, we simply use the previous entry's `vn_next` to jump over the removed one (except on the first entry, which has no previous, but is pointed at by the header's `sh_offset`).

#### To be done

**Two questions still make it a work in progress** (in addition to a firm code review, because I'm not familiar at all with ELF):
* how to handle removing of the last entry? Can I let a section with a size of 0, or should I remove it entirely?
* what to do with the versioned symbols that "loose" their version indication?\
  _this question will be handled in a separate comment, to avoid polluting this description once resolved_

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1768,8 +1768,12 @@ void ElfFile<ElfFileParamNames>::removeNeeded(const std::set<std::string> & libs
     char * strTab = (char *) fileContents->data() + rdi(shdrDynStr.sh_offset);
 
     auto dyn = (Elf_Dyn *)(fileContents->data() + rdi(shdrDynamic.sh_offset));
+    Elf_Dyn * vnNumDyn = 0;
     Elf_Dyn * last = dyn;
     for ( ; rdi(dyn->d_tag) != DT_NULL; dyn++) {
+        if (rdi(dyn->d_tag) == DT_VERNEEDNUM)
+            // Our stable location will be on last, not on dyn, in case of shifting.
+            vnNumDyn = last;
         if (rdi(dyn->d_tag) == DT_NEEDED) {
             char * name = strTab + rdi(dyn->d_un.d_val);
             if (libs.count(name)) {
@@ -1784,6 +1788,53 @@ void ElfFile<ElfFileParamNames>::removeNeeded(const std::set<std::string> & libs
     }
 
     memset(last, 0, sizeof(Elf_Dyn) * (dyn - last));
+
+    // If we've got a Verneed section AND it's advertised as having at least one entry.
+    unsigned int verNeedNum;
+    if (vnNumDyn && (verNeedNum = rdi(vnNumDyn->d_un.d_val)) > 0) {
+        Elf_Shdr & shdrVersionR = shdrs.at(getSectionIndex(".gnu.version_r"));
+        Elf_Shdr & shdrVersionRStrings = shdrs.at(rdi(shdrVersionR.sh_link));
+        char * verStrTab = (char *) fileContents->data() + rdi(shdrVersionRStrings.sh_offset);
+        debug("found .gnu.version_r with %i entries\n", verNeedNum);
+
+        auto need = (Elf_Verneed *)(fileContents->data() + rdi(shdrVersionR.sh_offset));
+        Elf64_Word * vnLastOffset = 0;
+        unsigned int vnRemovedCount = 0;
+        while (verNeedNum > 0) {
+            char * file = verStrTab + rdi(need->vn_file);
+            auto i = libs.find(file);
+            if (i != libs.end()) {
+                debug("skipping .gnu.version_r entry '%s'\n", file);
+                // If possible we'll just skip the entry by incrementing the previous' entry vn_next, letting our (now ghost) data in place.
+                // This will avoid breaking the binary in case of a strange, but technically possible, usage of the section where vn_next != sizeof(Elf_Verneed),
+                // where some arbitrary data has been interleaved between the Elf_Verneeds.
+                if (vnLastOffset)
+                    wri(*vnLastOffset, rdi(need->vn_next) ? rdi(*vnLastOffset) + rdi(need->vn_next) : 0);
+                else {
+                    // The only case where we trim is the first entry: its previous pointer is also the section's start (shdrVersionR.sh_offset).
+                    // @todo Not sure we let a working binary when removing the last entry (vn_next == 0).
+                    wri(shdrVersionR.sh_offset, rdi(shdrVersionR.sh_offset) + rdi(need->vn_next));
+                    wri(shdrVersionR.sh_size, rdi(shdrVersionR.sh_size) - rdi(need->vn_next));
+                }
+                ++vnRemovedCount;
+                changed = true;
+            } else {
+                debug("keeping .gnu.version_r entry '%s'\n", file);
+                vnLastOffset = &need->vn_next;
+            }
+            // vn_next == 0 officially ends the chained list (and would technically get us in an infinite loop).
+            if (rdi(need->vn_next) == 0)
+                break;
+            need = (Elf_Verneed *) (((char *) need) + rdi(need->vn_next));
+            --verNeedNum;
+        }
+        if (vnRemovedCount > 0) {
+            debug("shrinking count of versions from %d to %d\n", rdi(vnNumDyn->d_un.d_val), rdi(vnNumDyn->d_un.d_val) - vnRemovedCount);
+            wri(vnNumDyn->d_un.d_val, rdi(vnNumDyn->d_un.d_val) - vnRemovedCount);
+            wri(shdrVersionR.sh_info, rdi(shdrVersionR.sh_info) - vnRemovedCount);
+            replaceSection(".gnu.version_r", rdi(shdrVersionR.sh_size));
+        }
+    }
 
     this->rewriteSections();
 }


outtersg:
### Orphan symbols

#### Original problem

I started on FreeBSD 14.2, with a simple C++ program:
```c++
#include <iostream>

int main(int argc, char ** argv)
{
    std::cerr << "Plouf" << std::endl;
    return 0;
}
```
This program was first compiled with the system `clang++`, which linked it to `/lib/libc++.so.1` and `/lib/libcxxrt.so.1`.

After adding a custom `clang` toolchain (18.1.8 instead of the stock 18.1.6, but with its own `libc++.so.1` **and `libc++abi.so.1`**),\
the program crashed _after_ having run alright (`cerr` and `endl` correctly output), in its destructors part (`~sentry` tries to know if we are in exception handling with a call to `uncaught_exception()` that accesses something unitialized or overwritten).

I can resolve this crash either by:
1. `--add-needed libc++abi.so.1` which is inserted _before_ `libcxxrt.so.1` and thus gets called instead of `libcxxrt.so.1` over their common symbols
2. `--replace-needed libcxxrt.so.1 libc++abi.so.1` of course
3. `--remove-needed libcxxrt.so.1` thus letting `libc++.so.1` as the only needed lib (and thus the lookup for needed libs will cascade to the ones mentioned in `libc++.so.1`, including `libc++abi.so.1`)

However **`patchelf --remove-needed` _without this PR's patch_ corrupts the binary** for FreeBSD's loader (`ld-elf.so.1`), either when launching it or when inspecting it with `ldd`:

> ld-elf.so.1: /tmp/1.bin.sanscxxrt: Unexpected inconsistency: dependency libcxxrt.so.1 not found

`readelf` is a bit more robust and sees that the needed lib has been removed:

```diff
  0x0000000000000001 NEEDED               Shared library: [libc++.so.1]
- 0x0000000000000001 NEEDED               Shared library: [libcxxrt.so.1]
  0x0000000000000001 NEEDED               Shared library: [libm.so.5]
```

But the entry still in `.gnu.version_r` triggers the "Unexpected inconsistency":

```
Symbol table '.dynsym' contains 26 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
   […]
   7: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __cxa_begin_catch@CXXABI_1.3 (3)
   […]

Version symbol section (.gnu.version):
  000:   0 *local*         2 FBSD_1.7        1 *global*        1 *global*     
  004:   1 *global*        1 *global*        1 *global*        3 CXXABI_1.3   
  008:   1 *global*        3 CXXABI_1.3      4 GCC_3.0         1 *global*     
  00c:   5 GLIBCXX_3.4     1 *global*        1 *global*        1 *global*     
  010:   1 *global*        1 *global*        6 FBSD_1.0        1 *global*     
  014:   1 *global*        1 *global*        1 *global*        1 *global*     
  018:   1 *global*        3 CXXABI_1.3   

Version needed section (.gnu.version_r):
  0x0000 vn_version: 1 vn_file: libcxxrt.so.1 vn_cnt: 2
  0x0030   vna_name: CXXABI_1.3 vna_flags: 0 vna_other: 3
  0x0040   vna_name: GLIBCXX_3.4 vna_flags: 0 vna_other: 5
  0x0010 vn_version: 1 vn_file: libgcc_s.so.1 vn_cnt: 1
  0x0050   vna_name: GCC_3.0 vna_flags: 0 vna_other: 4
  0x0020 vn_version: 1 vn_file: libc.so.7 vn_cnt: 2
  0x0060   vna_name: FBSD_1.0 vna_flags: 0 vna_other: 6
  0x0070   vna_name: FBSD_1.7 vna_flags: 0 vna_other: 2
```

(here we see that the `vn_version`s are compacted at the start of the section (offsets 0x0, 0x10, 0x20), with the auxiliaries occupying the end of the section; and the `vna_other` field of the auxiliary section is the version name's ID in the `.gnu.version` section, which is appended to symbols expected in this library too in the .dynsym section)

#### Patching patchelf

With this PR's patch, we get our `.gnu.version_r` right:

```
Version needed section (.gnu.version_r):
  0x0000 vn_version: 1 vn_file: libgcc_s.so.1 vn_cnt: 1
  0x0040   vna_name: GCC_3.0 vna_flags: 0 vna_other: 4
  0x0010 vn_version: 1 vn_file: libc.so.7 vn_cnt: 2
  0x0050   vna_name: FBSD_1.0 vna_flags: 0 vna_other: 6
  0x0060   vna_name: FBSD_1.7 vna_flags: 0 vna_other: 2
```

(note the offsets shifted by -0x10 due to the "first entry gets trimmed" rule, and offsets 0x20 and 0x30 not being reused, because the ghost auxiliary entries CXXABI_1.3 and GLIBCXX_3.4 still sit there).

**The binary then works as expected** which was our primary goal.

However…

#### The internal orphan problem

Although the binary runs well, when `readelf`ing it, **the missing `.gnu.version_r` entry disrupts resolving in both `.gnu.version` and `.dynsym`**.

Both `binutils`'s `readelf` and the stock `readelf` of FreeBSD will fail to find the version node's name.

But binutils will fail gracefully, making it appear as an unversioned / corrupted symbol:
```
     […]
     6: […]
     7: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __cxa_begin_catch@@<corrupt>
     […]

Version symbols section '.gnu.version' contains 26 entries:
 Addr: 0x0000000000200578  Offset: 0x00000578  Link: 3 (.dynsym)
  000:   0 (*local*)       2 (FBSD_1.7)      1 (*global*)      1 (*global*)   
  004:   1 (*global*)      1 (*global*)      1 (*global*)      3              
  008:   1 (*global*)      3                 4 (GCC_3.0)       1 (*global*)   
  00c:   5                 1 (*global*)      1 (*global*)      1 (*global*)   
  010:   1 (*global*)      1 (*global*)      6 (FBSD_1.0)      1 (*global*)   
  014:   1 (*global*)      1 (*global*)      1 (*global*)      1 (*global*)   
  018:   1 (*global*)      3              
```
(note the `@@<corrupt>` in `.dynsym`, and the present-but-unnamed entries ID 3 in `.gnu.version`)

However FreeBSD's readelf will _stop reading those sections as soon as the first error is encountered_:
```
Symbol table '.dynsym' contains 26 entries:
     […]
     6: […]
     7: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __cxa_begin_catch
[and .dynsym dumping ends here]

Version symbol section (.gnu.version):
  000:   0 *local*         2 FBSD_1.7        1 *global*        1 *global*     
  004:   1 *global*        1 *global*        1 *global*     
[and .gnu.version dumping ends here]

[other sections continue without error]
```

(and it warns twice of "readelf: invalid versym version index 3", once per section it then stops dumping)

So…

### Question for this PR

This partial test (_limited to FreeBSD_) shows that, at least on 1 platform, we can rely on the runtime loader's robustness to handle dangerously formed binaries (fallbacking to an unversioned symbol when not finding the version node's name?),

Are there **use cases where this PR could prove useful as is** (and particularly: on NixOS, or with standard Linux distribs)\
or **should the removal of the versioned library _always_ un-version every related symbol too?**

satmandu:
@Ma27 Any thoughts on this?

PR #627, opened 2025-11-11
Drop i686-linux build

edolstra:
This already failed to evaluate due to pre-commit's dependency on dotnet-sdk, which is unsupported on i686-linux.

It also adds `nix flake check --all-systems --no-build` to CI to catch regressions like this.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,7 +22,6 @@
 
       supportedSystems = [
         "x86_64-linux"
-        "i686-linux"
         "aarch64-linux"
       ];
       forAllSystems = lib.genAttrs supportedSystems;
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -174,14 +173,12 @@
           constituents = [
             self.hydraJobs.tarball
             self.hydraJobs.build.x86_64-linux
-            self.hydraJobs.build.i686-linux
```
Mic92, line 177: should we not rather exclude pre-commit for 32-bit? I think patchelf in nixpkgs ist still needed for 32-bit libs.

Mic92: `pkgsi686Linux` still has quiet a few references.

edolstra: That's what I initially tried, but flake-parts seems to pull in pre-commit as well in the dev shell. We could just remove `devShells.i686-linux.default`...

PR #632, opened 2026-01-04
readme: document --add-rpath option

### LINKED ISSUE #586: Readme doesn't include the `--add-rpath` option
I, at least, did not even know about the `--add-rpath`, until I searched this repo for something similar. I had looked at the readme but didn't see this feature, but luckily searched. Of course, a simple `patchelf --help` also shows it often I'm looking at a project's repo to start with.
So, simple suggestion to put this in the readme (and anything else that might not be up to date), as usually the first point of contact for a prospective user.

conao3:
Fixes #586

This PR adds documentation for the `--add-rpath` option to the README.

The `--add-rpath` option was missing from the README, making it difficult for users to discover this feature. This option appends a directory to the existing RPATH, unlike `--set-rpath` which replaces it entirely.

---

This is a documentation-only change, so no regression test is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,16 @@ libraries.  In particular, it can do the following:
   $ patchelf --set-rpath /opt/my-libs/lib:/other-libs my-program
   ```
 
+* Add a directory to the `RPATH`:
+
+  ```console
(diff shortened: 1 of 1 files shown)


conao3:
thanks!

PR #635, opened 2026-01-28
Optimize stripping prefix in replaceNeeded

kugel-:
We can avoid resizing .dynstr if the user just requested to strip a prefix. More importanly, we can avoid relocating the section to the end of the binary which is almost always mandatory since .dynstr only ever grows.

This is a common use case when replacing fully-qualified needed entries (e.g. absolute build paths) with just its base name (to have the runtime linker do the search).

Thank you!


kugel-:
@edolstra @Ericson2314 Can one of you guys look at this please?

kugel-:
Replaced by #646

PR #641, opened 2026-05-21
Preserve pointer provenance during string table offset arithmetic

Flowdalic:
When calculating the address of the section header string table, the previous implementation cast the base data pointer to a size_t to perform integer arithmetic, and then cast the resulting integer back to a const char *.

This breaks standard C/C++ pointer provenance rules. Compilers rely on pointer arithmetic to track object origins, bounds, and aliasing. Losing this provenance chain by laundering the pointer through an integer type can lead to undefined behavior, as the compiler may no longer recognize the resulting pointer as safely pointing within the original allocation.

This patch resolves the issue by using standard pointer arithmetic instead of integer arithmetic. By adding the offset directly to the base pointer (fileContents->data() + shstrtabOffset) and only casting the final, derived pointer to the target type, the provenance chain remains perfectly intact.

Adhering to these provenance rules is strictly required on capability-based architectures like CHERI (e.g., CHERI RISC-V). On these systems, pointers are hardware capabilities containing bounds and validity tags. The integer cast stripped this metadata, resulting in an untagged, invalid pointer that caused a hardware trap (SIGSEGV) when reading the string table. This fix allows the compiler use correctly derived and valid capabilities.

Additionally, replace the `__builtin_add_overflow` check on the absolute memory address with a more straightforward bounds check against the total file size.


edolstra (approved):

Flowdalic:
Not sure what the workflow here is. Anything to do in my end?

PR #644, opened 2026-06-17
Fix collision detection for resized PHTs

### LINKED ISSUE #643: patchelf fails to detect it must relocate the PHT when it is already not at the beginning of the file
**Describe the bug**

Patching a shared library might require extending its program header table, which can then cause it to run out of its originally allocated space. After https://github.com/NixOS/patchelf/pull/544, `patchelf` handles this by detecting that growing the PHT in place would result in a collision with a section that won't be moved, in which case it instead moves the PHT to the end of the ELF file.

The logic checking for collisions is:

https://github.com/NixOS/patchelf/blob/71433faca34f1b89d23e4b12dfbeb5fec2866d13/src/patchelf.cc#L851-L860

where `phtSize` is set to:

https://github.com/NixOS/patchelf/blob/71433faca34f1b89d23e4b12dfbeb5fec2866d13/src/patchelf.cc#L836

That says that a section collides with the PHT if it's within the first `sizeof(Elf_Ehdr) + new_pht_size + padding_to_section_alignment` bytes of the file.

That's correct if the PHT is at the start of the file, but it's obviously wrong if the current PHT is somewhere else in the file! Any collision detection that never looks at `rdi(hdr()->e_phoff)` can't possibly be correct, because it's not looking in the right place.

**Steps To Reproduce**

Download and extract [libcrypto.so.1.1.tar.gz](https://github.com/user-attachments/files/29011912/libcrypto.so.1.1.tar.gz) to get `libcrypto.so.1.1` (which comes from the [`mysql_connector_python-8.0.28-cp310-cp310-manylinux1_x86_64.whl`](https://files.pythonhosted.org/packages/ea/1e/77be016984d903c633a22d76aff2bcdbb047aaaca4f63c015ec658f2519d/mysql_connector_python-8.0.28-cp310-cp310-manylinux1_x86_64.whl) wheel of https://pypi.org/project/mysql-connector-python/8.0.28/#files originally).

Then run:

```sh
patchelf --add-rpath '$ORIGIN:/does/not/exist' libcrypto.so.1.1
```

**Expected behavior**

patchelf should update the RUNPATH to `$ORIGIN:/does/not/exist`, but it instead fails with:
```
patchelf: cannot find section '.hash'
```

**`patchelf --version` output**

Reproduced on latest master, 71433faca34f1b89d23e4b12dfbeb5fec2866d13 (but `patchelf --version` just prints "patchelf"??)

**Additional context**

Per `readelf`, `libcrypto.so.1.1` has:
```
  Start of program headers:          2723840 (bytes into file)
```
and
```
Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 0x21597c 0x21597c R E 0x200000
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RW  0x8
  NOTE           0x000190 0x0000000000000190 0x0000000000000190 0x000024 0x000024 R   0x4
  GNU_EH_FRAME   0x1d1f38 0x00000000001d1f38 0x00000000001d1f38 0x00c184 0x00c184 R   0x4
  LOAD           0x216000 0x0000000000416000 0x0000000000416000 0x02c3d0 0x030868 RW  0x200000
  LOAD           0x299000 0x0000000000447000 0x0000000000447000 0x015c88 0x015c88 RW  0x1000
  DYNAMIC        0x299188 0x0000000000447188 0x0000000000447188 0x000200 0x000200 RW  0x8

 Section to Segment mapping:
  Segment Sections...
   00     .note.gnu.build-id .gnu.hash .dynsym .gnu.version .gnu.version_d .gnu.version_r .rela.dyn .rela.plt .init .plt .text .fini .rodata .eh_frame_hdr .eh_frame
   01
   02     .note.gnu.build-id
   03     .eh_frame_hdr
   04     .ctors .dtors .jcr .data.rel.ro .got .got.plt .data .bss
   05     .dynamic .dynstr
   06     .dynamic
```
The program header table is at 2723840 (0x299000), and the `.dynamic` section is at 0x299188, 392 bytes later. Since `sizeof(Elf64_Phdr)` == 56, there's enough room for exactly 392 / 56 == 7 program headers in the table - which is exactly how many program headers there are in the table, so the `.dynamic` section is directly after the program header table in the file.

The collision detection doesn't notice and thinks it can grow the PHT in-place, because it's looking for space at the start of the file instead of where the PHT currently is.

There seem to be two possible fixes here: changing the collision detection to look where the PHT currently is in order to grow the PHT in-place there, or keep the collision detection checking for space at the start of the file and relocate the PHT to the start instead of leaving it where it currently is if there's room at the start.

I'm not sure which of the two is preferable.

godlygeek:
Patching a shared library might require extending its program header
table, which can then cause it to run out of its originally allocated
space. After 484d349, patchelf handles this by detecting that growing
the PHT in place would result in a collision with a section that won't
be moved, in which case it instead moves the PHT to the end of the ELF
file.

The logic checking for collisions incorrectly assumes that the PHT is
originally at the start of the file, though. It checks every section
after the first to see if it overlaps with the PHT given its new size,
but it assumes that the PHT's original position is immediately after the
ELF header, `sizeof(Elf_Ehdr)` bytes into the file.

In other words, it's currently looking for collisions in the range:

    [0, roundUp(sizeof(Elf_Ehdr) + newPhtSize, sectionAlignment)]

This commit fixes it to check for collisions in the correct range:

    [hdr()->e_phoff, hdr()->e_phoff + roundUp(newPhtSize, sectionAlignment)]

Closes #643

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -833,7 +833,7 @@ void ElfFile<ElfFileParamNames>::rewriteSectionsLibrary()
     /* Compute the total space needed for the replaced sections, pessimistically
        assuming we're going to need one more to account for new PT_LOAD covering
        relocated PHDR */
-    off_t phtSize = roundUp((phdrs.size() + num_notes + 1) * sizeof(Elf_Phdr) + sizeof(Elf_Ehdr), sectionAlignment);
+    off_t phtSize = roundUp((phdrs.size() + num_notes + 1) * sizeof(Elf_Phdr), sectionAlignment);
     off_t shtSize = roundUp(rdi(hdr()->e_shnum) * rdi(hdr()->e_shentsize), sectionAlignment);
 
     /* Check if we can keep PHT at the beginning of the file.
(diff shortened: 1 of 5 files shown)


godlygeek:
It was surprisingly difficult to write a regression test for this. I'd be open to doing it a different way - either vendoring a `.so` where the PHT is in the middle of the ELF file instead of the start, or doing something like repeatedly calling `patchelf` to modify the same ELF file until _it_ relocates the PHT, and then making sure that the N+1th modification doesn't overwrite anything.

domenkozar:
Fixed in #652

PR #647, opened 2026-06-22
Add --build-resolution-cache to write a per-DSO library resolution note

domenkozar:
## What

Adds a `--build-resolution-cache` option that records, for each `DT_NEEDED`
entry of an ELF, where the dynamic loader would find it, in a `PT_NOTE`
(`.note.nixos.ldcache`, owner `NixOS`, type `0x63a86cb6`).

The descriptor is a sequence of NUL-terminated `(needed, path-list)` string
pairs, where each path-list element is either `=<absolute-path>` for a directly
resolved library or `?<dir>` for a directory the loader must still search
itself (used for `$ORIGIN`-relative and `glibc-hwcaps` directories).

## Why

On `/nix/store`-based systems every dependency lives in its own directory, so an
ELF's `DT_RUNPATH` lists one directory per dependency. To resolve each
`DT_NEEDED` soname the loader probes every `DT_RUNPATH` directory in turn,
producing a storm of failing `openat()`/`stat()` calls (roughly libraries times
directories) before each library is found. This measurably slows program
startup, especially on slow disks, network filesystems, and low power hardware.
See [NixOS/nixpkgs#481620](https://github.com/NixOS/nixpkgs/issues/481620).

A glibc loader that understands this note can resolve a binary's direct
dependencies straight from it, skipping the `DT_RUNPATH` walk entirely. The note
is consulted after `LD_LIBRARY_PATH` and before the `DT_RUNPATH` search, so
`LD_LIBRARY_PATH` / `LD_PRELOAD` / `/run/opengl-driver` overrides keep working;
an unpatched loader ignores the note.

This is the writer half; the loader half is a separate glibc patch.

## Behaviour

- The note is placed in its own page-aligned `PT_LOAD` covered by a `PT_NOTE`,
  with a matching `SHT_NOTE` section; the program and section header tables are
  relocated through the existing `rewriteSections()` path (handles both
  `ET_DYN` and the page-aligned `ET_EXEC` case).
- Resolves `DT_RUNPATH`, falling back to `DT_RPATH` when `DT_RUNPATH` is absent,
  mirroring the loader.
- Rebuilding after the run path changed fails loudly rather than keeping a stale
  note; an empty run path warns and writes nothing.
- Rejected together with `--force-rpath` (the cache only applies to
  `DT_RUNPATH`).

## Tests

`tests/build-resolution-cache.sh` (PIE, `DT_RPATH` fallback, stale-rebuild,
no-run-path, `LD_LIBRARY_PATH` override, idempotency, `--force-rpath`
rejection) and `tests/build-resolution-cache-no-pie.sh` (the non-PIE `ET_EXEC`
page-boundary case).

## Credit

Based on the per-DSO resolution-cache approach originally proposed by pennae in
[NixOS/nixpkgs#207893](https://github.com/NixOS/nixpkgs/pull/207893).

---

Disclosure: this change was developed with the assistance of Claude Code
(Claude Opus 4.8); the commit carries an `Assisted-by:` trailer, and the author
has reviewed and is accountable for it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kugel-:
If you know the path couldn't you just record the absolute path in `DT_NEEDED` (I've done that on other occasions).

domenkozar:
Yes, glibc honors a `/` in `DT_NEEDED` and opens that path directly, but that bypasses the search entirely, which is what the note is trying to avoid. Three reasons it's not equivalent:

- **Overrides break.** The note is consulted after `LD_LIBRARY_PATH` and before `DT_RUNPATH`, so `LD_LIBRARY_PATH` / `LD_PRELOAD` / `/run/opengl-driver` substitution keep working. An absolute `DT_NEEDED` skips the search, so those overrides no longer apply (on NixOS this pins the build-time driver instead of the runtime one).
- **No fallback.** A stale note just costs the normal RUNPATH walk, and an unpatched loader ignores it. A wrong absolute `DT_NEEDED` has no soname left to fall back on, so it fails hard.
- **No hwcaps.** The `?<dir>` entries let the loader still pick a `glibc-hwcaps` / `$ORIGIN` variant. A single absolute path can only point at the baseline.

Plus it keeps `DT_NEEDED` as the canonical soname list (tooling, `ldd`, RUNPATH shrinking) instead of baking store paths into it. The note is meant to be an advisory, overridable cache, not a hard pin.

Issue #639, opened 2026-05-17
DT_INIT not updated when --set-rpath relocates .init in lld-produced ELFs (SIGSEGV on dlopen)


Fedr:
## Summary

When `--set-rpath` causes patchelf to relocate `.init` (because the new RUNPATH string forces the program-header table to grow, and the original `.init` happened to sit immediately after the PHT), patchelf updates the **section header** for `.init` to its new location but **leaves `DT_INIT` in the dynamic section pointing at the old virtual address**. At runtime, `_dl_init` calls the address in `DT_INIT`, which is now overwritten by the newly-appended Phdr entry, and the process SIGSEGVs the moment the library is `dlopen`'d.

This appears to be a long-standing design gap in patchelf: I reproduced it identically on **0.15.0, 0.16.1, 0.17.2, and 0.18.0**. It overlaps with #446 (community pinning to 0.16.1) and #492 (alignment regression on 0.18) but is, I believe, a distinct underlying bug — `DT_INIT` simply isn't tracked when `.init` moves.

## Reproduction

Trigger requires an input ELF where (a) the program-header table is packed tight against `.note.gnu.build-id` / `.init` in the first PT_LOAD, and (b) `DT_INIT` is set. Modern `lld` produces exactly this layout by default on x86_64; `ld.bfd` typically leaves more padding and is not affected.

Minimal-ish recipe:

```bash
# Build a shared library with lld (any reasonably recent Clang).
cat > foo.c <<'EOF'
__attribute__((constructor)) static void init(void) {}
int foo(void) { return 42; }
EOF
clang -fuse-ld=lld -shared -fPIC -o libfoo.so foo.c

# Confirm the tight layout: PHT at 0x40..0x200, .note.gnu.build-id at 0x200,
# .init at 0x224, DT_INIT == 0x224.
readelf -d libfoo.so | grep -E '(INIT|RUNPATH|STRTAB)'
readelf -S libfoo.so | grep -E '\.(init|note\.gnu\.build-id|plt|dynstr)\b'

# Make patchelf grow the PHT — adding *any* RUNPATH that needs a new PT_LOAD
# triggers it. Real-world trigger: linuxdeploy adding $ORIGIN.
patchelf --set-rpath '$ORIGIN' libfoo.so

# .init is now at a high address (good!), but DT_INIT still says 0x224 (bug!).
readelf -d libfoo.so | grep INIT
readelf -S libfoo.so | grep '\.init\b'

# Verify the crash:
[...]

PR #652, opened 2026-06-24
Fix two PHT-relocation correctness bugs (#643, #639)

domenkozar:
Fixes two independent correctness bugs in the program-header-table (PHT) relocation logic, both of which currently block bumping patchelf in nixpkgs.

## #643 — in-place PHT growth assumes the PHT is right after the ELF header

`rewriteSectionsLibrary` decides whether a section collides with the grown PHT by comparing the section's file offset against `phtSize`, which only makes sense when the PHT sits at its canonical location right after the ELF header. When the PHT is elsewhere in the file — e.g. the OpenSSL `libcrypto.so.1.1` shipped in many manylinux wheels, whose PHT sits near the end — that check looks in the wrong place, the PHT is wrongly kept in place, and patching aborts with `cannot find section '.hash'`.

The fix detects the non-canonical PHT location up front (`e_phoff != sizeof(Elf_Ehdr)`) and relocates the PHT to the end of the file in that case.

**Verified** on the real `libcrypto.so.1.1` from the issue: before the fix patchelf errors `cannot find section '.hash'`; after, `--set-rpath` succeeds and the result is a structurally valid ELF.

A regression test is included (`tests/pht-relocation.sh`). The fixture is that real `libcrypto.so.1.1`, stripped and xz-compressed (decompressed at test time, skipped if `xz` is unavailable — mirroring the existing `short-first-segment.gz` fixture). I could not reproduce the corruption with a minimally-synthesized binary — it depends on the real multi-segment / many-section layout — so the fixture is larger (~730 KB) than the existing ones. Happy to swap it for a smaller trigger if anyone knows of one.

## #639 — `.dynamic` fixup never updated `DT_INIT` / `DT_FINI` / `DT_*_ARRAY`

The `.dynamic` fixup loop rewrites `DT_STRTAB`/`DT_SYMTAB`/`DT_HASH`/… to the new addresses of their sections, but never updated `DT_INIT`, `DT_FINI`, `DT_INIT_ARRAY`, `DT_FINI_ARRAY` or `DT_PREINIT_ARRAY`. When growing the PHT relocates `.init` (common with lld's tight layout), the loader then jumps to a stale `DT_INIT` and the process SIGSEGVs, typically on `dlopen`. The fix updates those entries from the relocated section addresses, like the others.

This fix is correct by construction and the full test suite stays green, but I was not able to synthesize the exact lld layout that relocates `.init`, so it has no dedicated regression test. Guidance on producing such a fixture welcome.

---

These are independent; I'm happy to split into two PRs if that's easier to review/merge.

Assisted-by: Claude Code (Claude Opus 4.8)

### DIFF
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1374,20 +1374,11 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
                 dyn->d_un.d_ptr = findSectionHeader(".gnu.version_r").sh_addr;
             else if (d_tag == DT_VERSYM)
                 dyn->d_un.d_ptr = findSectionHeader(".gnu.version").sh_addr;
-            /* The init/fini pointers reference the address of the
-               corresponding section, so they must be updated when those
-               sections are relocated (e.g. when growing the PHT pushes .init
-               to the end of the file); otherwise the dynamic loader jumps to a
-               stale address and the process crashes, typically at dlopen time.
-               See https://github.com/NixOS/patchelf/issues/639. */
-            else if (d_tag == DT_INIT) {
-                auto shdr = tryFindSectionHeader(".init");
-                if (shdr) dyn->d_un.d_ptr = (*shdr).get().sh_addr;
-            }
-            else if (d_tag == DT_FINI) {
-                auto shdr = tryFindSectionHeader(".fini");
-                if (shdr) dyn->d_un.d_ptr = (*shdr).get().sh_addr;
-            }
+            /* DT_{PREINIT,INIT,FINI}_ARRAY point at the corresponding
+               section's sh_addr and must follow it if relocated
+               (https://github.com/NixOS/patchelf/issues/639).
+               DT_INIT/DT_FINI are function addresses (-Wl,-init/-fini), not
+               section addresses, so leave them alone. */
             else if (d_tag == DT_INIT_ARRAY) {
                 auto shdr = tryFindSectionHeader(".init_array");
                 if (shdr) dyn->d_un.d_ptr = (*shdr).get().sh_addr;
--- a/src/patchelf.cc
+++ b/src/patchelf.cc
@@ -1374,6 +1374,32 @@ void ElfFile<ElfFileParamNames>::rewriteHeaders(Elf_Addr phdrAddress)
                 dyn->d_un.d_ptr = findSectionHeader(".gnu.version_r").sh_addr;
             else if (d_tag == DT_VERSYM)
                 dyn->d_un.d_ptr = findSectionHeader(".gnu.version").sh_addr;
+            /* The init/fini pointers reference the address of the
+               corresponding section, so they must be updated when those
+               sections are relocated (e.g. when growing the PHT pushes .init
+               to the end of the file); otherwise the dynamic loader jumps to a
+               stale address and the process crashes, typically at dlopen time.
+               See https://github.com/NixOS/patchelf/issues/639. */
+            else if (d_tag == DT_INIT) {
+                auto shdr = tryFindSectionHeader(".init");
+                if (shdr) dyn->d_un.d_ptr = (*shdr).get().sh_addr;
+            }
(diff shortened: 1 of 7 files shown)


Fedr:
Confirming this addresses a real downstream case — I filed #639 after hitting it while packaging MeshLib's Python bindings (an lld-built `libpybind11nonlimitedapi_*.so` run through `patchelf --set-rpath` by `linuxdeploy`).

I recovered the actual triggering binary from one of our daily builds (clang/lld 21). It has exactly the tight layout that's hard to synthesize — `.init` packed immediately after an 8-entry PHT:

```
PHDR                       off 0x40   (8 entries, ends at 0x200)
[ 1] .note.gnu.build-id    0x200
[ 2] .init                 0x224      <- right after the PHT
[ 3] .plt                  0x240
[ 4] .text                 0xff0
DT_INIT = 0x224
```

Reproduction with **patchelf 0.18.0** (the released line `linuxdeploy` bundles) — appending *any* path to the existing RUNPATH is enough:

```
$ patchelf --set-rpath '<existing>:$ORIGIN/x' shim.so
$ readelf -SW shim.so | grep ' .init '   # .init moved to 0x32028
$ readelf -dW shim.so | grep '(INIT)'    # DT_INIT still 0x224  <- stale
```

→ `_dl_init` calls the stale `0x224` (now overwritten by the appended Phdr) and SIGSEGVs on `dlopen`. Built from this PR, `DT_INIT` tracks `.init` and the library loads cleanly.

**One caveat if you want it as a regression test:** on current `master`, patchelf relocates the whole PHT to end-of-file for this binary instead of growing it in place, so `.init` isn't relocated and it's handled correctly *even without this PR* — the corruption only appears with the 0.18.0-era in-place growth. So this `.so` confirms #639 is real and that `master` no longer mangles it, but it wouldn't fail as a test without the `DT_INIT` change; you'd need an input that still takes the in-place-growth path to exercise that code directly. The binary is attached here: https://github.com/user-attachments/files/29332263/libpybind11nonlimitedapi_meshlib_3.12.so.zip

domenkozar:
Thanks for the detailed confirmation and for recovering the real binary.

You're right, and it goes a bit deeper than this one binary. In `rewriteSectionsLibrary` the in place growth path bails to full PHT relocation as soon as it hits a non replaceable section in range, and `.init` is `SHT_PROGBITS` so it always triggers that. The executable path also refuses to move `SHT_PROGBITS` sections. So on current master `.init` never gets relocated in place, which is why your `.so` loads fine with or without the `DT_INIT` change.

Given that, I'll keep the #639 fix as defensive correctness (it mirrors the existing `DT_STRTAB`/`DT_SYMTAB` fixups) rather than add a regression test that can't actually fail. The #643 path is covered directly by the test in this PR.

Thanks again, the binary is a useful end to end sanity check.

domenkozar:
Restructured this PR to build on @godlygeek's #644 rather than duplicate it:

- The #643 fix is now his two commits (`Ensure PHT collision detection checks the right spot` + `Fix collision detection for resized PHTs`), cherry-picked with authorship preserved. His version is more precise than my original: it corrects the collision-detection range to `[e_phoff, e_phoff + phtSize)` and drops the bogus `+ sizeof(Elf_Ehdr)` term, so it only relocates the PHT when there's a genuine collision instead of giving up in-place growth for any non-canonical PHT location.
- His synthetic `pht-collision` test replaces the 748 KB compressed libcrypto fixture I had added. It's deterministic, needs no blob, and handles ELF32/64 and endianness.
- My remaining commit is just the #639 fix (update DT_INIT/DT_FINI/DT_INIT_ARRAY/DT_FINI_ARRAY/DT_PREINIT_ARRAY when those sections move), which #644 doesn't cover.

Verified locally: `pht-collision.sh` passes, and @Fedr's real lld binary patches cleanly into a valid ELF. If preferred, #644 can land on its own and I'll rebase the #639 commit on top.

Mic92:
@domenkozar feel free to drop a release

